PR #10, opened 2020-05-05
Validate that algorithms are tar files.

manthey:
Validating that they are actually docker images is harder.  Specifically, that should be done asynchronously, but Django doesn't provide for async validators (yet -- they apparently will be added in a future version).

As a future work, whenever we save an algorithm, we should start an async job to inspect the docker image and store known metadata back to the algorithm record so that we have more information to present to the user.


manthey:
Thanks.  I figured as soon as PRs start having any chance of conflicting with each other, we should start actually reviewing them.

PR #13, opened 2020-05-05
Run algorithm and scoring jobs when queued.

manthey:
As soon as the status is changed to queued, these jobs will be scheduled to run.


manthey:
@banesullivan I've based this on your dev-ui branch.

PR #21, opened 2020-05-08
Added overall_score and result_type

JellyWang-3:
I am not sure if I rebased correctly and I might did squash wrong.
overall_score and result_type columns seem to work locally.


banesullivan:
Hm, it looks like something went awry with your rebase... also, we should avoid committing any experimental migrations to the database and just have one per PR with a squash and merge.

@JellyWang-3, Since this looks like a smaller diff (which is great), I'd recommend:

1. delete your local database and recreate it while checked out on `master`
2. copy/past that database to a safe place and give it a name like `master-db-b4-<feature-branch>`
3. extract your relevant diffs into a new branch off of `master`
4. then only run `makemigrations` once to avoid committing intermediate migrations that were created during dev/testing.

@manthey and I were talking the other day about how to avoid these kinds of experimental migrations. I am very inexperienced with this, but it sounds like the easiest route *for now* is for everyone to have safe/external copy of the DB based on the `master` branch so that we can easily throw away migrations made during dev/testing and only commit a single migration per PR.

We may want to consult others more familiar with Django about this.

manthey:
@JellyWang-3 and I discussed how to fix the rebase.

For simple migrations, you can actual unmigrate a django database if the migrations aren't too complex.  You can do, for instance:
```
manage.py migrate core 0011
```
and it will move the database to state 0011.

I recommend that for any given PR, we try to have just one migration file.  Therefore, once you think things are ready, you can migrate to the last migration in master, delete all of the locally made migrations, the remake a single migration with `makemigrations` and migrate to that.

Obviously, if you add a column, you'll lose values from that column when you migrate backwards, but it will keep the history and merges clean.

banesullivan:
Ah, I didn't realize you could unmigrate the DB - that makes the process much easier than what I suggested above.

manthey:
> Ah, I didn't realize you could unmigrate the DB - that makes the process much easier than what I suggested above.

There are some complex migrations that aren't reversible without additional code (for instance, you could manually aggregate columns as part of a migration), but generally if you can create the necessary migrations only using `makemigrations`, they are reversible.

JellyWang-3:
Thank you guys for helping me out with my PR.
I have cleaned the branch and migration. I will keep your suggestions on my mind when I work on the code.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
manthey, line 1: Since this is a vs specific file, it should be removed from the PR.  You can add `.vscode` to the `.gitignore` file to prevent it from being checked in in the future.

core/models.py
```diff
@@ -100,6 +99,11 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse('algorithm-detail', kwargs={'creator': str(self.creator), 'pk': self.pk})
 
+    def __str__(self):
```
manthey, line 102: Somehow these lines have been added when they shouldn't have been.

core/models.py
```diff
@@ -166,6 +172,31 @@ def post_save(self, created, *args, **kwargs):
...
+        tasks.run_algorithm.delay(self.id)
+
+    def post_save(self, created, *args, **kwargs):
+        if not created and kwargs.get('update_fields') and 'status' not in kwargs.get('update_fields'):
+            return
+        if self.status == self.Status.QUEUED:
+            self.run_algorithm()
+        # We may want to implement canceling here
+
+
```
manthey, line 199: This was somehow duplicated.

core/models.py
```diff
@@ -225,5 +256,8 @@ def post_save_score_job(sender, instance, *args, **kwargs):
 class ScoreResult(models.Model):
     score_job = models.ForeignKey(ScoreJob, on_delete=models.CASCADE, blank=True, null=True)
     created = models.DateTimeField(default=timezone.now)
+    #  setting the upload directory and file name
     data = models.FileField(upload_to='scores')
     log = models.FileField(upload_to='scores_logs', null=True, blank=True)
+    overall_score = models.FloatField(null=True, blank=True, validators=[MaxValueValidator(1.0), MinValueValidator(0.0)])
+    result_type = models.CharField(max_length=10, null=True, blank=True)
```
manthey, line 263: We want to have the possible values for `result_type` defined in the model.  For instance:
```
    class ResultTypes(models.TextChoices):
        SIMPLE = 'simple', _('Direct value')
        ROC = 'roc', _('Receiver Operating Characteristic')
```
and then
```
result_type = models.CharField(max_length=10, choices=Status.choices, null=True, blank=True)
```

.vscode/settings.json
JellyWang-3: Thanks. I will do that immediately.

core/tasks.py
```diff
@@ -154,13 +156,18 @@ def _run_scoring(score_job):
...
             score_result.data.save(
                 'score_job_%s.dat' % score_job.id, open(output_path, 'rb'))
             score_result.log.save(
                 'score_job_%s_log.dat' % score_job.id, open(stderr_path, 'rb'))
+            # one line code to extract the float from data file
+            score_result.overall_score = (float([line.strip() for line in score_result.data][0]))
+            # depends on the data type the return type varies
+            result_type = Enum('result_type', 'simple roc')
+            # for now, always return 'simple' as result type
+            score_result.result_type = (list(result_type)[0].name)
```
manthey, line 170: I think we want a function to process this so to conceptually separate it.  In the future, it will be more complex.  For instance:
```
def _overall_score_and_result_type(datafile):
    # In the future, inspect the data to determine the result type.  For now, just extract a float from the data file
    result_type = ScoreResult.ResultType.SIMPLE
    overall_score = float(score_result.data.readline())
    return overall_score, result_type
```
and in this function
```
        score_result.overall_score, score_result.result_type = _overall_score_and_result_type(score_result.data)
```

core/tasks.py
```diff
@@ -154,13 +154,15 @@ def _run_scoring(score_job):
...
-            # Store result
             score_result = ScoreResult(
                 score_job=score_job)
             score_result.data.save(
                 'score_job_%s.dat' % score_job.id, open(output_path, 'rb'))
             score_result.log.save(
                 'score_job_%s_log.dat' % score_job.id, open(stderr_path, 'rb'))
+            score_result.overall_score, score_result.result_type = _overall_score_and_result_type(score_result.data)
+            # score_result.overall_score = float(0.92)
+            # score_result.result_type = 'SIMPLE'
```
manthey, line 165: Can you remove these lines?

JellyWang-3: yes.

core/migrations/0013_auto_20200511_1458.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.0.6 on 2020-05-11 14:58
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0012_auto_20200511_1449'),
```
manthey, line 10: Because of the temporary rename, this won't be correct.  It should be `'0012_auto_20200507_1615'`

JellyWang-3: I fixed it. Sorry about that

PR #24, opened 2020-05-12
Add a simple algorithm that emits a log.

manthey:


manthey:
I have docker images of all of the items that are in the utils directory if anyone needs them for testing.

PR #25, opened 2020-05-12
log preview for algorithm and score

JellyWang-3:
1. Display message so user know "This only show the end of the log"
2. log preview for both algorithm and score results with only 100 kb


core/models.py
```diff
@@ -179,6 +179,8 @@ class AlgorithmResult(models.Model):
     created = models.DateTimeField(default=timezone.now)
     data = models.FileField(upload_to='results')
     log = models.FileField(upload_to='results_logs', null=True, blank=True)
+    log_message = models.TextField(default='This only show the end of the log')
```
manthey, line 182: We don't actually want to add these to the models as fields, since that adds them to the database.  Similarly, we don't want the task to generate this.  The task just saves the log file.

Rather, there should just be a function in the `AdminModel` for `AlgorithmResultAdmin` and `ScoreResultAdmin` (much like what is currently in master, but generalized for the two models as a common function).

I'd expect that there shouldn't need to be a migration for this, as the core models shouldn't change.

JellyWang-3: So both log_message and log_preview should not be added to the models?
Instead, they should be a generalized function that takes the file and return corresponding outputs?

manthey: Neither should be added to the models.

For logs, we just want to show the end of the log (your log_message).

If the *data* is known to be a text mime-type, then we will want to show the beginning of the data.  Currently, we don't check or store the data mime type; we, ideally, would have different previews based on the type.  For, now if we showed the beginning of the data as text, that is still useful for testing.

JellyWang-3: For log_message, it should be a message telling the user that we are only showing the end of the log or it should show the end of the log file(the last line of the log file)?
And for log_preview, show the beginning of the data as text?

manthey: For logs, we want to show the last ~100kB of the log file as text.  If it isn't the complete file, we should mention that.  For instance, Circle CI does this by printing:
```
Your output is too large to display in the browser.
Only the last 400000 characters are displayed.
```
For our use case this should probably be:
```
The log output is too large to display in the browser.
Only the last 100000 characters are displayed.
```
where that 100000 should be a variable in the function so we aren't duplicating it in multiple places.

For generality, we probably can have a single utility function, `def _text_preview(filefield, maxlen=100000, showEnd=True):`, where for logs it would call it with `filefield=<model>.log`, and for data it would call it with `filefield=<model>.data, showEnd=False)`.

core/admin.py
```diff
@@ -5,6 +5,8 @@
 from . import models
 from . import tasks
 
+import os
```
manthey, line 8: We don't (yet) have a style rule for this, but usually we put imports from external and system packages before local packages.  Can move this to just below the django imports?

core/admin.py
```diff
@@ -65,11 +67,19 @@ def algorithm(self, obj):
     def dataset(self, obj):
         return obj.algorithm_job.dataset
 
+    # how to combine the function?
     def log_preview(self, obj):
```
manthey, line 71: This function can be a module level function.  We don't use `self` within the function, so within each AdminModel there would be a much shorter function (e.g.,
```
  def log_preview(self, obj):
    return _text_preview(obj.log)
```
and then the module level would be renamed to `_text_preview` and take the FileField object rather than the model.

core/admin.py
```diff
@@ -65,11 +67,19 @@ def algorithm(self, obj):
...
+        maxlen = 10000
+        log_message = 'The log output is too large to display in the browser.\n'
+        log_message += ('Only the last %s characters are displayed.\n \n' % (maxlen))
         if obj.log:
-            log = '\n'.join(obj.log.readlines())
+            log = '\n'.join(obj.log.open('rt').readlines())
```
manthey, line 76: This is inefficient and won't work if there is binary data in the log.  Rather, we want to do something like:
```
if obj.log:
  prefix_message = ''
  with obj.log.open('rb') as datafile:
    try:
      datafile.seek(-maxlen, os.SEEK_END)
    except OSError as exc:
      if exc.errno != 22:  
        raise  # reraise exceptions except for trying to seek before the beginning
      prefix_message = ('The output is too large to display in the browser.\n' +
       f'Only the last {maxlen} characters are displayed.\n\n')
    message = datafile.read().decode(errors='replace')
    return prefix_message + message
```

core/migrations/0014_auto_20200507_1756.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.0.6 on 2020-05-07 17:56
```
manthey, line 1: If you rebase or merge master, you shouldn't have any migrations or changes to models or tasks.

core/admin.py
JellyWang-3: yes

PR #26, opened 2020-05-12
Configure Django for production deployment

brianhelba:


brianhelba:
@manthey @banesullivan I'm merging this now. There are a few small follow-up PRs I'll make immediately, and I had to disable the unit tests, until we configure the third-party services to work during testing, but this should be a significant step towards a production-ready config.

PR #29, opened 2020-05-14
Log preview

JellyWang-3:
Github history clean version of log preview.

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -1,9 +1,39 @@
 from django.contrib import admin
 from django_admin_display import admin_display
 from django.utils.safestring import mark_safe
+import os
 
 from . import models
 from . import tasks
+
+
+# should take the FileField object rather than the model.
+# a module for log preview
+def _text_preview(log_file):
+    # max file size for display, currently 100kb
+    maxlen = 10000
+    prefix_message = 'The output is too large to display in the browser.\n'
+    prefix_message += ('Only the last %s characters are displayed.\n \n' % (maxlen))
+    if log_file:
+        with log_file.open('rb') as datafile:
+            if len(datafile) > 0:
+                try:
+                    datafile.seek(-maxlen, os.SEEK_END)
+                except OSError as exc:
+                    if exc.errno != 22:
+                        # reraise exceptions except for trying to seek before the beginning
+                        raise
+                message = datafile.read().decode(errors='replace')
+                if len(log_file) < maxlen:
+                    # different prefix depends on the output size?
+                    prefix_message = 'Log output: \n'
+                    return prefix_message + message
+                else:
+                    return prefix_message + message
+            else:
+                return 'Log is empty'
+    else:
+        return 'No log file to display'
 
 
 @admin_display(short_description='Run algorithm')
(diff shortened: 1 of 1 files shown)


core/admin.py
```diff
@@ -1,11 +1,41 @@
...
 from django.utils.safestring import mark_safe
+import os
 
 from . import models
 from . import tasks
 
 
+# should take the FileField object rather than the model.
+# a module for log preview
+def _text_preview(log_file):
```
manthey, line 12: It is more common to document functions using a doc string.  That is, rather than a comment above the function, this would be:
```
def _text_preview(log_file):
    """
    Return the text of a file if it is short or the last portion of it if it is long.

    :params: log_file A FileField to read text from.
    """
```

Alternately, since we are using a new enough version of Python, you can use type decorators (e.g.,
```
def _text_preview(log_file: FileField):
```

core/admin.py
```diff
@@ -1,11 +1,41 @@
...
+    # max file size for display, currently 100kb
+    maxlen = 10000
```
manthey, line 14: This should match the comment (either 100kb or 10kb).

core/admin.py
```diff
@@ -1,11 +1,41 @@
...
+                try:
+                    datafile.seek(-maxlen, os.SEEK_END)
+                except OSError as exc:
+                    if exc.errno != 22:
+                        # reraise exceptions except for trying to seek before the beginning
+                        raise
+                message = datafile.read().decode(errors='replace')
+                if len(log_file) < maxlen:
+                    # different prefix depends on the output size?
+                    prefix_message = 'Log output: \n'
```
manthey, line 29: I don't think we need to prefix anything when it is the full output.

core/admin.py
JellyWang-3: I got a "NameError: name 'FileField' is not defined" so I changed some code to work around it and lint test. But I am not sure if I made the correct fix.

core/admin.py
```diff
@@ -1,10 +1,44 @@
 from django.contrib import admin
 from django_admin_display import admin_display
 from django.utils.safestring import mark_safe
+from django import forms as file_field
```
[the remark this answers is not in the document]
JellyWang-3: oh, thanks

core/admin.py
```diff
@@ -1,11 +1,43 @@
...
+def _text_preview(log_file: FileField):
+    """
+    Return the text of a file if it is short or the last portion of it if it is long.
+
+    :params: log_file A FileField to read text from.
+    """
+    # max file size for display, currently 10kb
+    maxlen = 10000
+    prefix_message = 'The output is too large to display in the browser.\n'
+    prefix_message += ('Only the last %s characters are displayed.\n \n' % (maxlen))
```
manthey, line 20: Since prefix_message is only used by one if-else branch, can we move its definition to that location?

Also, you can do this with a single multi-line format string 
```
    prefix_message = f"""The output is too large to display in the browser.
Only the last {maxlen} characters are displayed.

"""
```

JellyWang-3: Yes, I will do that.

core/admin.py
```diff
@@ -1,11 +1,44 @@
...
+                    if exc.errno != 22:
+                        # reraise exceptions except for trying to seek before the beginning
+                        raise
+                message = datafile.read().decode(errors='replace')
+                if len(log_file) < maxlen:
+                    return message
+                else:
+                    prefix_message = f"""The output is too large to display in the browser.
+                        Only the last {maxlen} characters are displayed.
+                        """
```
manthey, line 34: In triple-quote string formatting, white-space is preserved, so we want don't want spaces on the left for subsequent lines (since it will indent that line in the display).

To match the formatting you had earlier, we also want a blank line before the closing triple quotes.

You can see how it looks in the application by temporarily setting the `maxlen` threshold to a small number.

PR #31, opened 2020-05-14
Align the log output

JellyWang-3:
For over 10kb, I do noticed that the new line between "The output is too large to display in the browser."
and "Only the last {maxlen} characters are displayed" disappeared, should I add it back or it doesn't matter?


manthey:
I think to get the new line in the maxlen message to appear, you'll need to run the prefix message through whatever filter django uses that converts line feeds to `<br/>`.  It would be useful to know how to do that.

PR #33, opened 2020-05-15
mime type check for algorithm results  

JellyWang-3:
Done: 
Able to display mime type:
Selection_050
1. check if the file is empty: if yes return None(or should I let it return "inode/x-empty"?)
2. get the mimetype of the file and store both results with comma seperated if it has 2

Do we need to return file_mimetype and ontent_mimetype separately? 
Should I format the returned mime type?
Todo: 
Double check on the lint test and clean up comments after code is reviewed


core/tasks.py
```diff
@@ -238,3 +243,31 @@ def _overall_score_and_result_type(datafile):
...
+
+def _get_mimetype(file_path):
+    mimetype = magic.from_file(file_path, mime=True)
+    # check if mimetype has something useful
+    # empty file returns inode/x-empty
+    if mimetype == 'inode/x-empty':
+        # if no mime type, it can be null(None)
+        return None
+    # Unsure about how to get file and content mime type seperately
+    mime_type = mimetype.split('/')
```
manthey, line 256: The minetype is expected to have a slash in it.  We don't want to split it.

core/tasks.py
```diff
@@ -238,3 +243,31 @@ def _overall_score_and_result_type(datafile):
...
+    # file_mimetype = mime_type[0]
+    content_mimetype = mime_type[1]
+    """
+    check if the file is zipped since true
+    or false of uncompress has no effect on non-zip file
+    """
+    if content_mimetype == 'zip':
```
manthey, line 263: There are many compression format that uncompress can handle (zip, gzip, bz2, probably others).  I would generically always check the `uncompress=True` type since we don't want to enumerate the compression formats that the magic library handles.

core/tasks.py
JellyWang-3: I thought the slash is to separate the file mime type and content mime type. So the whole thing together is a mime type for the file?

core/tasks.py
JellyWang-3: Oh I did not take that into consideration. I will change it

core/tasks.py
manthey: Yes, the whole thing is the mime type.  Pedantically, it is a type and subtype, but they are almost always used as a single value.

core/tasks.py
```diff
@@ -238,3 +243,24 @@ def _overall_score_and_result_type(datafile):
...
+    unzipped_file = magic.Magic(mime=True, uncompress=True)
+    unzipped_file_mimetype = unzipped_file.from_file(file_path)
+    zipped_file = magic.Magic(mime=True, uncompress=False)
+    zipped_file_mimetype = zipped_file.from_file(file_path)
+    # if the file is unzipped
+    if unzipped_file_mimetype == zipped_file_mimetype:
+        return unzipped_file_mimetype
+    else:
+        # seperate unzipped and zipped mimetype by comma
+        file_mimetype = ('%s,%s' % (unzipped_file_mimetype, zipped_file_mimetype))
```
manthey, line 264: There is no need to wrap this in parenthesis (e.g., `file_mimetype = '%s,%s' % (unzipped_file_mimetype, zipped_file_mimetype)` is better).

core/tasks.py
```diff
@@ -238,3 +243,24 @@ def _overall_score_and_result_type(datafile):
...
+    mimetype = magic.from_file(file_path, mime=True)
+    # check if mimetype has something useful
+    # empty file returns inode/x-empty
+    if mimetype == 'inode/x-empty':
+        # if no mime type, it can be null(None)
+        return None
+    unzipped_file = magic.Magic(mime=True, uncompress=True)
+    unzipped_file_mimetype = unzipped_file.from_file(file_path)
+    zipped_file = magic.Magic(mime=True, uncompress=False)
+    zipped_file_mimetype = zipped_file.from_file(file_path)
```
manthey, line 258: We already have the value of the possibly compressed file in `mimetype` from line 249.

core/models.py
```diff
@@ -215,6 +215,16 @@ class AlgorithmResult(models.Model):
...
+    data_mimetype = models.TextField(null=True, blank=True)
+    """
+    use the magic package on the output_path and store
+    the resulting mime_type, if any, to the new field.
+    magic.from_file(<filename>, mime=True)
+    should allow the field to be null (None
+    if magic.from_file(<filename>, mime=True, uncompress=True)
+    returns a different value that without the uncompress flag.
+    either store both mimetypes (perhaps comma-separated)
+    """
```
manthey, line 227: Remove this docstring, please.  If you feel the `data_mimetype` field needs commenting, it should just be an ordinary comment and probably precede the definition.

core/tasks.py
```diff
@@ -238,3 +243,24 @@ def _overall_score_and_result_type(datafile):
...
+
+def _get_mimetype(file_path):
+    mimetype = magic.from_file(file_path, mime=True)
+    # check if mimetype has something useful
+    # empty file returns inode/x-empty
+    if mimetype == 'inode/x-empty':
+        # if no mime type, it can be null(None)
+        return None
+    unzipped_file = magic.Magic(mime=True, uncompress=True)
+    unzipped_file_mimetype = unzipped_file.from_file(file_path)
```
manthey, line 256: For clarity, I would use a different variable name that `unzipped_file`, as the object in the variable is actually a Magic class.  Perhaps `uncompressed_magic` and `uncompressed_mimetype` would be more accurate names.

PR #40, opened 2020-05-19
Override default textarea to text input

JellyWang-3:
Override docker_image_id and data_mimetype in formfield
I cannot find a way to override default textarea for fail_reason when it is empty.

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -4,6 +4,8 @@ from django.utils.safestring import mark_safe
 from django.utils.html import escape
 from django.template.defaultfilters import linebreaksbr
 from django.db.models import FileField
+# override the default textarea
+from django.forms import TextInput
 
 import os
 
@@ -42,6 +44,16 @@ def _text_preview(log_file: FileField):
     else:
         return 'No log file to display'
 
+# def formfield_for_dbfield(self, db_field, **kwargs):
+#     formfield = super().formfield_for_dbfield(db_field, **kwargs)
+#     if db_field.name == 'fail_reason':
+#         # check if return object value is empty
+#         if kwargs['fail_reason'] == None:
+#             return formfield
+#         else:
+#             # if not empty do not override
+#             return
+
 
 @admin_display(short_description='Run algorithm')
 def run_algorithm(modeladmin, request, queryset):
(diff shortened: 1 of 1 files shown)


core/admin.py
```diff
@@ -68,6 +69,13 @@ def data_link(self, obj):
 
     data_link.allow_tags = True
 
+    # overrride default textfield model
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        formfield = super().formfield_for_dbfield(db_field, **kwargs)
+        if db_field.name == 'docker_image_id':
+            formfield.widget = TextInput(attrs=formfield.widget.attrs)
+            return formfield
```
manthey, line 77: This line is over indented, which will prevent many fields from appearing.

core/admin.py
```diff
@@ -150,6 +165,12 @@ def data_link(self, obj):
 
     data_link.allow_tags = True
 
+    def formfield_for_dbfield(self, db_field, **kwargs):
```
manthey, line 168: The `ScoreAlgorithmAdmin` class is nearly identical to the `AlgorithmAdmin` class.  Although it doesn't have to do directly with this PR other than that you are adding to it, we can replace the entire `ScoreAlgorithmAdmin` class with:

```
@admin.register(models.ScoreAlgorithm)
class ScoreAlgorithmAdmin(AlgorithmAdmin):
    list_display = ('__str__', 'task', 'creator', 'created', 'active', 'data_link')
```
to reduce code duplication.

core/admin.py
```diff
@@ -138,7 +153,7 @@ def data_link(self, obj):
 
 
 @admin.register(models.ScoreAlgorithm)
-class ScoreAlgorithmAdmin(admin.ModelAdmin):
+class ScoreAlgorithmAdmin(AlgorithmAdmin):
     list_display = ('__str__', 'task', 'creator', 'created', 'active', 'data_link')
     readonly_fields = ('data_link',)
```
manthey, line 158: Because we are now subclassing `AlgorithmAdim`, we can remove lines 143-151 (the entire class just contains the  `list_display` definition).

PR #46, opened 2020-05-21
This allows using a docker-based worker.

manthey:
Some of our tasks run docker commands.  To do this within a docker container, it is necessary to mount the docker command and the docker socket.  This uses the default locations of each of these as deployed in ubuntu.

Further, one of the virtues of developing with docker is network separation.  However, since the ports were specified in the docker-compose.yml file, this wasn't full possible -- the docker development environment has no need to expose these to the host computer, and if the host already had one these ports in use it would fail.  This moves the port exposure to `docker-compose.override.native.yml`, and the instructions for using the native python have changed appropriately.

### DIFF
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -28,7 +28,7 @@ def _field_file_to_local_path(field_file: FieldFile) -> Generator[Path, None, No
     with field_file.open('rb'):
         file_obj: File = field_file.file
 
-        if isinstance(file_obj, S3Boto3StorageFile):
+        if not Path(file_obj.name).exists() or isinstance(file_obj, S3Boto3StorageFile):
             field_file_basename = PurePath(field_file.name).name
             with tempfile.NamedTemporaryFile('wb', suffix=field_file_basename) as dest_stream:
                 shutil.copyfileobj(file_obj, dest_stream)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,10 +6,12 @@ services:
       POSTGRES_DB: rgd
       POSTGRES_PASSWORD: postgres
     volumes:
-      - /var/lib/postgresql/data
+      - "dbdata:/var/lib/postgresql/data"
 
   rabbitmq:
     image: rabbitmq:latest
+    volumes:
+      - mqdata:/var/lib/rabbitmq
 
   minio:
     image: minio/minio
@@ -18,4 +20,9 @@ services:
       MINIO_ACCESS_KEY: minioAdminAccessKey
       MINIO_SECRET_KEY: minioAdminSecretKey
     volumes:
-      - /data
+      - fsdata:/data
+
+volumes:
+  dbdata:
+  fsdata:
+  mqdata:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,13 +7,9 @@ services:
       POSTGRES_PASSWORD: postgres
     volumes:
       - /var/lib/postgresql/data
-    ports:
-      - 5432:5432
 
   rabbitmq:
     image: rabbitmq:latest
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -22,6 +20,9 @@ services:
       MINIO_ACCESS_KEY: minioAdminAccessKey
       MINIO_SECRET_KEY: minioAdminSecretKey
     volumes:
-      - /data
-    ports:
-      - 9000:9000
+      - fsdata:/data
+
+volumes:
+  dbdata:
```
brianhelba, line 26: Why the need for named volumes?

core/tasks.py
```diff
@@ -28,7 +28,7 @@ def _field_file_to_local_path(field_file: FieldFile) -> Generator[Path, None, No
     with field_file.open('rb'):
         file_obj: File = field_file.file
 
-        if isinstance(file_obj, S3Boto3StorageFile):
+        if not Path(file_obj.name).exists() or isinstance(file_obj, S3Boto3StorageFile):
```
brianhelba, line 31: When does this case occur? When is the file not on disk but also not a `S3Boto3StorageFile`?

docker-compose.yml
manthey: If I do `docker-compose up`, it creates volumes with long hash-like names.  If I then do `docker-compose down` and `docker-compose up`, it does **not** reuse the old names.  Rather, it creates **new** hash-like names.  This makes the data functionally non-persistent.  Further, each time you do an up and down, you end up creating more volumes that use storage space that are never automatically cleaned up in any manner.  In addition to being functionally non-persistent, it is also resource consuming.  Using named volumes works around this.

If there is a better way, I would be happy to know it.

core/tasks.py
manthey: This happens in the development setup.  The file is not an `S3Boto3StorageFile`, and the path is relative to some root but not the current cwd, so the path appears to not exist.

You can see this by uploading an algorithm to the workflow; when it tries to validate the docker, the file path is not complete enough to find the file.

docker-compose.yml
brianhelba: `docker-compose stop` (not `down`) should be used to stop the containers. It persists and reuses the containers and volumes for the next `up`. For that reason, `stop` is the command suggested in the README.

Even if you have named volumes, if you want to cleanly destroy everything with `down`, you need [to pass the `-v` / `--volumes` flag](https://docs.docker.com/compose/reference/down/), which will remove both named and anonymous volumes.

Unless we need named volumes, I think it's a better practice to let Docker Compose manage volume names anonymously, which ensures there are never any name conflicts.

core/tasks.py
brianhelba: Yes, good point. I realize since we're using `MinioMediaStorage` in development, the file does not exist, but also is not a `S3Boto3StorageFile`.

This is a good general fix for right now, thanks.

docker-compose.yml
manthey: Using `docker-compose down` should **NOT** destroy persistent data.  This is a terrible default behavior.  Using named volumes fixes this.

Sure, stopping and starting is nice, but there are some things that stopping does allow -- for instance if you want to change volume mounts (for instance, to bind mount in a directory to easily access additional code within the container).

brianhelba: Following our conversation, I agree with your fundamental point.

I think we should prefix these volume names with `rgd-` to prevent conflicts with other projects which may create volumes with the same names.

manthey: The names are already prefixed with "resonantgeodata_" by docker-compose.

PR #50, opened 2020-05-26
algo result data preview

JellyWang-3:


core/admin.py
```diff
@@ -119,7 +145,10 @@ def dataset(self, obj):
         return obj.algorithm_job.dataset
 
     def log_preview(self, obj):
-        return _text_preview(obj.log)
+        return _text_preview(obj.log, obj.data_mimetype, True)
```
manthey, line 148: We don't want to pass the data_mimetype for the log, since it only applies to the data.  We can assume logs are supposed to be readable.

This applies to the other class that has a `log_preview` function, too.

core/admin.py
```diff
@@ -12,36 +12,46 @@
 from . import tasks
 
 
-def _text_preview(log_file: FileField):
+def _text_preview(target_file: FileField, mimetype, show_end):
```
manthey, line 15: We might want default values for `mimetype` and `show_end`, so callers don't have to specify them if they are the default, e.g. `def _text_preview(target_file: FileField, mimetype: string=None, show_end: bool=True):`.

core/admin.py
```diff
@@ -12,36 +12,46 @@
...
     """
     Return the text of a file if it is short or the last portion of it if it is long.
 
-    :params: log_file A FileField to read text from.
+    params: target_file A FileField to read text from.
+            mimetype Log or Data's mimetype, score result don't have one
+            show_end Determine if showing the first/last portion for preview
     """
+    mimetype_check = mimetype.startswith('text/')
```
manthey, line 23: If we allow `mimetype` to be `None`, this will change to `mimetype_check = mimetype is not None and mimetype.startswith('text/')`

manthey: We should return right away when the mimetype check fails.  We don't need to open the file or check if it exists.

core/admin.py
```diff
@@ -12,36 +12,46 @@
...
     maxlen = 10000
-    if log_file:
-        with log_file.open('rb') as datafile:
+    if target_file:
+        with target_file.open('rb') as datafile:
             if len(datafile) > 0:
-                try:
+                if show_end:
+                    # read and only display from the end for log filess                    display_message = 'last'
                     datafile.seek(-maxlen, os.SEEK_END)
```
manthey, line 31: We need to guard against seeking before the beginning, so this will be `datafile.seek(max(0, len(target_file)-maxlen))`.  Since we are now seeking from the current position and the file has just been opened, we don't have to specify a seek reference (though you could specify `os.SEEK_SET`).

core/admin.py
```diff
@@ -12,36 +12,46 @@
...
     # max file size for display, currently 10kb
     maxlen = 10000
-    if log_file:
-        with log_file.open('rb') as datafile:
+    if target_file:
+        with target_file.open('rb') as datafile:
             if len(datafile) > 0:
-                try:
+                if show_end:
+                    # read and only display from the end for log filess                    display_message = 'last'
```
manthey, line 30: It looks like `display_message = 'last'` has been combined with the previous line.

core/admin.py
```diff
@@ -12,36 +12,46 @@
...
                     datafile.seek(-maxlen, os.SEEK_END)
-                except OSError as exc:
-                    if exc.errno != 22:
-                        # reraise exceptions except for trying to seek before the beginning
-                        raise
-                message = datafile.read().decode(errors='replace')
-                if len(log_file) < maxlen:
+                    message = datafile.read().decode(errors='replace')
```
manthey, line 32: We can have a single read call for both conditions of `show_end` (that is, `datafile.read(maxlen).decode(errors='replace')` is valid for both conditions), so the read can be outside of the `if`/`else`.

core/admin.py
```diff
@@ -12,36 +12,46 @@
...
+                else:
+                    # only show message if it's log or mime type is text
+                    if mimetype is None or mimetype_check:
+                        display_message = 'first'
+                        datafile.seek(max(0, len(datafile) - maxlen))
```
manthey, line 37: We don't want to seek when showing the beginning.

JellyWang-3:
I'll fix them.

PR #51, opened 2020-05-28
Allow downloading files from dockerized minio.

manthey:
The download url for dockerized minio includes the docker internal network name and port (e.g., starting with `http://minio:9000/`).  DNS should not be expected to resolve this, not should forwarding that port be needed for development.

This adds an api endpoint (`/api/download/<snake_case_model>/<id>/<field>`) which will download any FieldFile record in a core model.  The download is provided with the internal name of the file and, if known, a stored mimetype.

The download_file function is probably too open (though we should only load models based on some user authorization).  The explicit check for `//minio:` is ugly and brittle.

**Note**: I view this as the *wrong* way to fix downloading from minio, but I'm not sure of a better way to do it that actually works in our typical development environment without adding a proxy server to the mix.  Specifically, the development download code path is now *special* and not what we are using in production.  Further, downloads are sent through the main django process rather than being deferred to the storage system -- since the `minio` hostname is not exposed to the host computer (nor external computers accessing the host computer), some alteration of the url appears to be necessary no matter what.

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -44,16 +44,13 @@ def _text_preview(log_file: FileField):
         return 'No log file to display'
 
 
-@admin_display(short_description='Run algorithm')
-def run_algorithm(modeladmin, request, queryset):
-    for algorithm_job in queryset:
-        tasks.run_algorithm.delay(algorithm_job.id)
-
-
-@admin_display(short_description='Run scoring')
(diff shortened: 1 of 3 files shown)


manthey:
I'm going to merge this to master.  When we think of a better way to do this so that development is more like production, we can revert changes as appropriate.

Issue #17, opened 2020-05-07
GeoData API Technical Requirements


aashish24:

aashish24:
- We will need to support temporal component
- 3D mesh data needs to be supported in the future

manthey:
For reference: https://github.com/OpenGeoscience/girder_geospatial

banesullivan:
I have been looking into what GeoDjango can handle and it's pretty compelling - I've been taking some notes and want to try some things out before I report too much else here. In brief, GeoDjango is a Django interface to PostGIS with support for vectorized data (shapefiles) and raster imagery where these data are typically static (no time variance) all implemented over GDAL.

------

In the meantime, I have been thinking a lot about how we can approach handling temporal data when working with rasters and how we need to be aware that down the road we will also need to support 3D data. A significant challenge arises when implementing a "model" to handle temporally varying remote sensing (RS) data: temporally varying data is most often also spatially varying in RS, however in my experience temporal 3D data is typically spatially static and only the data/values that live on the mesh are varying. 

For example, take any path and row of Landsat 8 imagery that you're interested in (e.g. Path 34 and Row 32 for Golden, CO) and query available data through time for that tile. You'll notice that the path shifts through time though our ROI (Golden) is in each of the tiles at different parts of the image:

![Screen Shot 2020-05-08 at 4 07 31 PM](https://user-images.githubusercontent.com/22067021/81445654-4ab06080-9147-11ea-8778-f8c7e4a99b98.png)


Currently, we can work with this by simply storing each raster in the DB individually with a timestamp and querying all rasters through time based on a specific area of interest. This is what it seems just about everyone else does when dealing with time-varying RS data. And it is an easy, somewhat lightweight solution as rasters are just images - the geometry to define them is nothing more than coordinates of the origin, pixel sizes, and pixel extents. So it is unlikely these elements will be repeated across your spatiotemporally varying data.

However, this sort of "model" might conflict with an approach to handle temporally varying 3D data where the geometry/extent of them does not change (spatially static). In this case, the dataset itself may contain a time-varying field but the geometries will not change through time. For example, a 3D model of contaminant transport in the subsurface where the mesh/spatially domain is constant, but the field values change through time:

*<😞, I can't immediately find a cool GIF for this>*

This makes me wonder what would be the best way to approach implementing a model framework that can handle both time-varying 2D data (rasters) and time-varying 3D data (physical models, etc). So far where I am at is:

[...]

PR #56, opened 2020-06-04
🚧 First pass at geodata app

### LINKED ISSUE #64: Create raster data model that can store multispectral data

### LINKED ISSUE #66: Create a ETL workflow to ingest raster data

banesullivan:
Addresses #17 

**WORK IN PROGRESS:** The app is not currently running with the site. No DB migrations are included. Whenever we do merge, **we should squash all commits during the merge** - I'd like to keep the history on this branch for now though.

## Models

Resolves #66 and resolves #64 

Note: we no longer use the `GDALRaster` entry or GeoDjango's `RasterField` but instead, define all the same fields on a custom model along with a GeoDjango `PolygonField`. This gives us a ton of flexibility in where we can store raster data (especially once they become large) and we can still use the `PolygonField` to leverage GeoDjango's spatial indexing for spatial searching and operations. When it comes to cropping the images, etc., we will need to leverage `large_image`.

These models implement a `SpatialEntry` abstract model which can be either raster data or geometry data. Each of these points to an archive/file of the data so that we are not storing large images twice. All geometry/shapes/features *are* stored in the DB.

A `Dataset` is a model for aggregating `SpatialEntry`s together so that users can group rasters and geometry together for projects.

`BandMetaEntry` keeps track of all the meta info specific to bands within a `RasterEntry` - again, no data is stored here, just metadata.

`ConvertedRasterFile` is a basic implementation of a model that will allow us to export rasters to common formats and store them in our file system for on-demand access. Down the road, we can create conditions to automatically clean these up.

geodata

### Meta class in `geometry`

This in an (arguably useless) wrapper classes that list some of the different meta-data properties of `GeometryCollection`. I added this to show what data is available from the GeoDjango interface that we do not need to repeat in our models.

Perhaps we can use these meta classes when creating views.

## Data ingestion routines (ETL)

See #66 

## Added dependencies and purpose

- `rasterio`: reading raster meta info - a lightweight wrapper around GDAL to choose the correct drivers to load the raster images
- `fiona`: File IO for vector geometry (shapefiles)
- `shapely`: Conversion of shapes read by `fiona` to an input `GEOSGeometry` can accept in GeoDjango

## Notes
- [x] Test against sample raster data:
  - [x] Landsat8 GeoTIFF rasters
  - [x] IARPA S2A `.jp2` rasters
  - [x] LAIR data (`NITF` rasters)

## Quirks

- I implemented a `PostSaveEventMixin` that has a set of shared functionality for calling post save tasks. It reduces repeated code but has the potential to introduce recursive imports. This should be reviewed and possibly changed.
- I did not implement a routine to compute stats on the bands because that would require loading each band into memory which could be expensive. Likewise for generating a thumbnail... we could use `rasterio` to subsample the raster and only read in a small amount of data to create a thumbnail though.


banesullivan:
@manthey, would you please take a look at this when you get a chance? The description above has been updated.

I have tested this against a few different example raster datasets and all seems to be working as long as GDAL has a driver for the given raster.

Mostly, I would like you to check that the models implemented here have a good structure and that I am using the best relationships for our use cases.

geodata/models/common.py
```diff
@@ -0,0 +1,73 @@
...
+class ModifiableEntry(models.Model):
+    """A base class for models that need to track modified datetimes."""
+
+    modified = models.DateTimeField(editable=False, help_text='The last time this entry was saved.')
+    created = models.DateTimeField(editable=False, help_text='When this was added to the database.')
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
```
manthey, line 26: In Girder, we track the modified date but also the user who did the modification.  Similarly, we track the user who created the object.  Do we want to do that here as well?  This would add a creator field and a modifier field.

geodata/models/common.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class SpatialEntry(ModifiableEntry):
+    """Common model to all geospatial data entries."""
+
+    name = models.CharField(max_length=100, blank=True, null=True)
+    # An optional description field in case the user needs to add context
+    description = models.TextField(blank=True, null=True)
+    # Datetime of creation for the dataset
+    acquisition_date = models.DateTimeField(default=timezone.now)
```
manthey, line 40: I think we want the acquisition_date to be nullable and default to null.  If we do a search based on acquisition time, it could be misleading to have a recent time here when the data could be much older.

geodata/models/geometry/base.py
```diff
@@ -0,0 +1,51 @@
...
+from ..mixins import PostSaveEventMixin
+from ... import tasks
+
+
+def validate_zip_extension(value):
+    import os
+    from django.core.exceptions import ValidationError
+
+    ext = os.path.splitext(value.name)[1]  # [0] returns path+filename
+    valid_extensions = [
```
manthey, line 17: Instead of using extension, we could also validate via file contents (using the python magic library, for instance, to ensure it is a zip file).

Is there something else that takes advantage of this being a zip file, or could we more broadly also accept tar files?

geodata/models/raster/base.py
```diff
@@ -0,0 +1,96 @@
...
+    # i.e. these are populated programatically by the ETL routine
+
+    raster_file = models.OneToOneField(RasterFile, null=True, on_delete=models.CASCADE)
+    # thumbnail = models.ImageField(blank=True, upload_to='thumbnails')
+
+    # This can be used with GeoDjango's geographic database functions for spatial indexing
+    footprint = models.PolygonField(srid=DB_SRID)
+
+    # Raster fields
+    crs = models.CharField(max_length=100)  # TODO: proj4 strings can get verbose... is 100 enough
```
manthey, line 56: 100 isn't enough.  I found one that is 236 characters in the EPSG database.

geodata/models/common.py
banesullivan: This is a good idea - I will add `creator` and `modifier` user fields.

Should we keep a history of the modifiers or is just updating the field with the laster user to modify it okay?

geodata/models/raster/base.py
```diff
@@ -0,0 +1,96 @@
...
+    origin = fields.ArrayField(models.FloatField(), size=2)
+    extent = fields.ArrayField(models.FloatField(), size=4)
+    resolution = fields.ArrayField(models.FloatField(), size=2)  # AKA scale
```
manthey, line 59: Do you know what units the resolution is in?

geodata/models/raster/base.py
```diff
@@ -0,0 +1,96 @@
...
+        blank=True,
+        help_text='Automatically retreived from raster but can be overwritten.',
+    )
+    dtype = models.CharField(max_length=10)
+    max = models.FloatField(null=True)
+    min = models.FloatField(null=True)
+    mean = models.FloatField(null=True)
+    std = models.FloatField(null=True)
+    nodata_value = models.FloatField(null=True)
+    parent_raster = models.ForeignKey(RasterEntry, on_delete=models.CASCADE)
```
manthey, line 83: When present, we may want to store the interpretation, too.  Multispectral data often just has "undefined", though.

geodata/models/raster/reader.py
```diff
@@ -0,0 +1,128 @@
...
+                    pass
+                try:
+                    band_meta.nodata_value = src.nodatavals[i]
+                except IndexError:
+                    pass
+                try:
+                    band_meta.dtype = src.dtypes[i]
+                except IndexError:
+                    pass
+                # TODO: Compute statistics: min, max, mean, std
```
manthey, line 114: Many files store min, max, mean, std for each band.  These don't always match the data.

If we add code to compute these, it might be useful to actually generate a histogram of values, too (though then we have to decide on binning, etc.).

geodata/models/raster/reader.py
```diff
@@ -0,0 +1,128 @@
...
+    def _read_files(self):
+        """Load raster data files and create a ``RasterEntry``.
+
+        This function will need to handle a lot of scenarios for how
+        the raster file is passed and where it is stored.
+
+        """
+        # Fetch the raster file this Layer corresponds to
+        self.rfe = RasterFile.objects.get(id=self.model_id)
+        file_path = self.rfe.raster_file.name
```
manthey, line 44: Because we are using S3/minio, we can't just do this.  Rather, we should use the `_field_file_to_local_path` context function (and that function should get moved to some common place to avoid duplication in core).  If that function is available, then this would be
```
with _field_file_to_local_path(self.rfe.raster_file) as file_path:
   <all following code>
```

geodata/models/raster/reader.py
```diff
@@ -0,0 +1,128 @@
...
+
+        logger.info(f'The raster file path: {file_path}')
+
+        raster_querry = RasterEntry.objects.filter(raster_file=self.rfe)
+        if len(raster_querry) < 1:
+            self.raster_entry = RasterEntry()
```
manthey, line 50: When we create a new RasterEntry, we should set the name to the raster file's name be default.

geodata/models/raster/reader.py
```diff
@@ -0,0 +1,128 @@
...
+        the raster file is passed and where it is stored.
+
+        """
+        # Fetch the raster file this Layer corresponds to
+        self.rfe = RasterFile.objects.get(id=self.model_id)
+        file_path = self.rfe.raster_file.name
+
+        logger.info(f'The raster file path: {file_path}')
+
+        raster_querry = RasterEntry.objects.filter(raster_file=self.rfe)
```
manthey, line 48: Can you change this to `raster_query`?

geodata/models/raster/reader.py
manthey: I pushed a commit that does this.

geodata/models/raster/base.py
banesullivan: What about using a `TextField` for these PROJ strings?

geodata/models/raster/base.py
banesullivan: The units would be defined by the CRS - i.e. we'll know the coordinate system from the PROJ string in the `crs` field and these values only make sense in that context.

geodata/models/raster/base.py
banesullivan: What do you mean by "interpretation"? Like red, blue, green, IR, etc? Should I make this another `TextField`?

manthey: Yes.  gdal, at least, has a enum of possible values (gray, palette, red, green, blue, etc).  Sadly, it doesn't include IR.  I could see it being a useful filter.

geodata/models/raster/base.py
manthey: Sure.

geodata/models/common.py
banesullivan: I have to make `ModifiableEntry` non-abstract to do this.

banesullivan: Also, access the user in the `save` method that I override isn't clear. I don't think the request is passed, so we may have to use a middleware plugin like this: https://blndxp.wordpress.com/2016/03/04/django-get-current-user-anywhere-in-your-code-using-a-middleware/

geodata/models/geometry/base.py
banesullivan: It could be any type of archive - zip/tar/whatever. I just chose zip for simplicity.

I will update this yo use the python magic library to verify the archive and also accept tar

banesullivan: I updated this to check the mime type. @manthey, can you take a look to see if this is what we want?

geodata/models/raster/reader.py
banesullivan: For now, I will update this PR to use the values stored directly in the raster. Any further implementation warrants a new issue and subsequent PR in my opinion as we'll need t put a bit of thought into how to compute these stats and bins if the values in the raster are not valid.

geodata/models/common.py
banesullivan: I added these fields with null available. For now, the user isn't saved as I'm not sure what is the best way to access the current user

banesullivan: I no longer set these variables in the `save()` event due to bad practice.

banesullivan:
@manthey: at this point, I think I've addressed everything you brought up and met the goals for this PR.

I cannot currently verify that the reader routine for shapefiles is working as I am having issues with GDAL when running the site locally (the setup code for the paths doesn't appear to be working on Mac). But from what I can tell, the routine should work - it'd be good to verify this though.

manthey:
@banesullivan With some small changes, this is working in my local docker deployment for a few raster files and the two zip files you shared.

One thing I'd like to see (in a different PR) is, if no name is specified for a raster file or geometry archive, to use the base name of the file for the name.

I'll approve this, and if you think my last commit is okay, go ahead and merge.  Any issues we discover can be dealt with on a future PR.

banesullivan:
Great, thanks for checking over this and for the fixes!

PR #58, opened 2020-06-05
Update development instructions

banesullivan:
@brianhelba, you may want to quickly double-check this.

The install instructions needed some corrections.

It had been a while since I had tried to launch the site directly (I have been working in a separate toy project to build the `geodata` app) and I ran into very minor issues which are fixed here.

`.env.docker-compose-native.sh` wasn't compatible with my system (I'm using `zsh` on Mac) so I changed it to export each variable rather than iterating over a string of variables (I don't know how that originally worked, that format was strange to me. Now it is more explicit)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,12 @@ but allows developers to run the Python code on their native system.
 
 ### Initial Setup
 1. Run `./dev/init-minio.sh`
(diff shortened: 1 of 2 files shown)


banesullivan:
@manthey and @brianhelba, I can't run the app locally without these changes so I am going to merge into #56

PR #61, opened 2020-06-09
Use PostGIS Docker image for development

brianhelba:


brianhelba:
See https://github.com/postgis/docker-postgis#postgispostgis for description

manthey:
@brianhelba In order to actually use the gis extensions, I modified the Dockerfile to install GDAL and the settings to use postgis.

If you are okay with these changes, merge away.

PR #68, opened 2020-06-11
Use a GDAL wheel to supply libgdal and libgeos_c.

manthey:


banesullivan:
@manthey, is that GDAL wheel built against `libcurl`?

> `/vsis3/` is a file system handler that allows on-the-fly random reading of (primarily non-public) files available in AWS S3 buckets, without prior download of the entire file. It requires GDAL to be built against libcurl.

https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files-random-reading

manthey:
Yes.  This gdal definitely works with vsis3.

PR #70, opened 2020-06-16
Add flower container and routing.

### LINKED ISSUE #53: Add an endpoint to get status of workers
It would be useful to know if we have any workers associated with the system, and, if so, if they are busy.

Specifically, this could probably be something like is done in girder_worker (see https://github.com/girder/girder_worker/blob/master/girder_worker/girder_plugin/api/worker.py#L40-L55).  For this purpose, the celery app can be reached via `from rgd import celery_app`.

Ideally, this let's us determine the following conditions:
- The broker is unavailable 
- There are no workers
- The number of idle workers
- The number of busy workers (and, ideally, what they are busy doing)

In the future, we may have multiple worker pools (for instance, for GPU and non-GPU tasks), so this will probably change exactly what gets reported in the future.

manthey:
This still needs to be added to the Heroku deployment.

Resolves #53.

### DIFF
--- a/core/templates/admin/add_flower.html
+++ b/core/templates/admin/add_flower.html
@@ -1,0 +1,20 @@
+{% extends "admin/index.html" %}
+{% load i18n %}
+
+{% block content %}
+  <div id="content-main">
+  {{ block.super }}
+
+  <div class="module current-app">
+    <table>
+      <caption>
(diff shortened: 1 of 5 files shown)


manthey:
@mattrix27 If this works for you, please approve it and then I'll merge.

PR #76, opened 2020-06-18
Drf core

mattrix27:
Added Viewsets and Serializers to expose core in Swagger via drf-yasg

endpoint in `/swagger`

You should be able to look up everything in core using the GET requests

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -1,35 +1,23 @@
 import logging
 import os
-import re
 
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
-from django.db.models.fields.files import FieldFile, FileField
 from django.db.models.fields import AutoField
+from django.db.models.fields.files import FieldFile, FileField
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, render
 from django.utils import timezone
 from django.utils.encoding import smart_str
 from django.views.decorators.http import require_http_methods
 from django.views.generic import CreateView, DeleteView, DetailView
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import viewsets
 from rest_framework.decorators import api_view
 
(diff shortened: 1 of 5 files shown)


manthey:
Since all of these ViewSets and all of the serializers and all of the registers routes repeat the same basic code for each model in core, it would be great if we can generalize this to avoid code repetition.

For instance, instead of defining a class for each viewset, we could have a loop like this for registering routes:
```
for ser in [
    serializers.AlgorithmSerializer,
    <list other serializers here>
]:
    model = ser.Meta.model
    name = model.__name__
    viewsetClass = type(name + 'ViewSet', (viewsets.ModelViewSet, ), {
        'queryset': model.objects.all(),
        'serializer_class': ser,
        'filter_backends': [DjangoFilterBackend],
        'filterset_fields': views.get_filter_fields(model),
    })
    router.register('api/%s' % (name.lower()), viewsetClass)
```

This could be taken even further: you could walk through the attributes of the core/model module and, for each model, create a serializer and then a viewset, and then register it.  However, I could needing a custom serializer for a model.

core/views.py
```diff
@@ -144,3 +148,82 @@ def download_file(request, model, id, field):
...
+    Return a list of all filterable fields of Model.
+
+    -Takes: Model type
+    -Returns: A list of fields as string (excluding ID and file uploading)
+    """
+    model_fields = model._meta.get_fields()
+    fields = []
+    for field in model_fields:
+        res = str(field).split('.')
+        if res[1] == model.__name__ and not isinstance(field, (FileField, AutoField)):
```
manthey, line 164: We probably DO want to serialize the file fields in some manner (though this can be a future PR).  Specifically, we want to at least mention that the file exists so we could use the download endpoint to fetch it (and, ultimately, also have a way to redirect to get it directly from S3 when appropriate).  For PUT/POST operations, we want to expose the ability to upload new files, too.

Issue #80, opened 2020-06-23
Use django-s3-file-field


manthey:
This should allow slow uploads in Heroku

manthey:
Added in #84.

PR #84, opened 2020-06-30
Use django-s3-file-field.

manthey:
This avoids direct uploads in the development environment.  It would be better if the django-s3-file-field package had a more sensible way to control this.


manthey:
Resolves #80.

PR #85, opened 2020-06-30
Drf core config

mattrix27:
Clean up Viewset production and created custom parser to handle manytomany fields w/file uploads


core/urls.py
```diff
@@ -1,25 +1,75 @@
...
+        result = super().parse(stream, media_type=media_type, parser_context=parser_context)
+
+        data = {}
+        arrays = {}
+
+        model = None
+        if parser_context and 'view' in parser_context:
+            model = parser_context['view'].get_serializer_class().Meta.model
+        for key, value in result.data.items():
+            """
```
manthey, line 29: Use a comment, not a docstring.

core/urls.py
```diff
@@ -1,25 +1,75 @@
...
+                Handle ManytoMany field data, parses lists of comma-separated integers that might be quoted.
+                eg. "1,2"
+            """
+            if hasattr(model, key) and isinstance(
```
manthey, line 33: This `if` can be somewhat simplified: `if isinstance(getattr(model, key, None), fields.related_descriptors.ManyToManyDescriptor):`, since `None` won't be an instance of `ManyToManyDescriptor`.

core/urls.py
```diff
@@ -1,25 +1,75 @@
...
+                getattr(model, key), fields.related_descriptors.ManyToManyDescriptor
+            ):
+                print(value)
```
manthey, line 36: Get rid of debug lines, please.

core/urls.py
```diff
@@ -1,25 +1,75 @@
...
+                arrays[key] = [x for x in vals.split(',')]
+            else:
+                data[key] = value
+
+        print(data)
+        print(arrays)
+        qdict = QueryDict('', mutable=True)
+        qdict.update(data)
+
+        for key in arrays:
```
manthey, line 47: Rather than collecting information in `data` and `arrays`, this could build the new `QueryDict` directly.  That is, instead of have two loops, you could so something like:
```
        qdict = QueryDict('', mutable=True)
        for key, value in result.data.items():
            # Handle ManytoMany field data: parse lists of comma-separated
            # integers that might be quoted.  e.g. "1,2"
            if isinstance(
                getattr(model, key, None), fields.related_descriptors.ManyToManyDescriptor
            ):
                for val in value.split(','):
                    qdict.update({key, val.strip('"')})
            else:
                qdict.update({key, value})
```

core/urls.py
mattrix27: Done

core/urls.py
mattrix27: Done!

core/urls.py
mattrix27: Done!

core/urls.py
mattrix27: Done!

manthey:
@mattrix27 Can you address the linting errors?

PR #86, opened 2020-06-30
Set default name based on file upload

### LINKED ISSUE #83: use base name of uploaded files by default in GeoData app
if no name is specified for a raster file or geometry archive, use the base name of the file for the name.

from https://github.com/ResonantGeoData/ResonantGeoData/pull/56#issuecomment-649766048

banesullivan:
Resolve #83


geodata/models/geometry/base.py
```diff
@@ -46,6 +47,11 @@ class GeometryArchive(ModifiableEntry, PostSaveEventMixin):
     geometry_entry = models.OneToOneField(GeometryEntry, null=True, on_delete=models.DO_NOTHING)
     failure_reason = models.TextField(null=True, blank=True)
 
+    def save(self, *args, **kwargs):
+        if self.name is None or len(self.name) < 1:
```
manthey, line 51: I'm curious why you chose `len(self.name) < 1` rather than `not len(self.name)`.

banesullivan: Simply out of habit. Since this is `len` of a `str`, it should be `not len(self.name)`.

I'm used to always doing the `< 1` check because there's no guarantee that `len` will return a positive integer - well, only on custom data types where `__len__` may be overridden. This is totally irrelevant here because we're dealing with a string, but I just developed that habit when checking `len` and it stuck

manthey: Since this is a string, the whole `if` statement could be changed to just: `if not self.name:`, since both `None` and a zero-length string will be falsy.

banesullivan: Ah yep! Much cleaner

PR #89, opened 2020-07-07
Add a basic search endpoint.

manthey:


manthey:
image

manthey:
@banesullivan SpatialEntry is now non-abstract, so search works generically on both RasterEntry and GeometryEntry.  Do we want a search parameter to only return one or the other?

banesullivan:
This is awesome having general search endpoint that returns all `SpatialEntry`s whose footprint matches the criterion! It may be good to add two additional, separate endpoints that use this search function for only Geometry and Raster rather than an option to filter?:

- `geodata/near_point`
- `geodata/raster/near_point`
- `geodata/geometry/near_point`

Or perhaps we could just add the option to the first endpoint and have the second two map to that?

manthey:
I like the idea of specific endpoints for raster and geometry.

There are two approaches to endpoints -- having many very specific endpoints or a having fewer endpoints with lots of options.  The virtue of many specific endpoints is that each endpoint is easier to test and understand.  The virtue of fewer endpoints with many options is that there is better chance that they cover unanticipated use cases.

I'm leaning toward have many specific endpoints, but I think we'll also end up having one very general endpoint.

banesullivan:
> I'm leaning toward have many specific endpoints, but I think we'll also end up having one very general endpoint.

I too am leaning in favor of many specific endpoints. I think having many specific endpoints each with a clear intention will yield us a more stable and easy to maintain API. Having a single endpoint with many, many options would be nice from a user perspective though.

I have a feeling we will have tons of endpoint feature requests no matter what and it'll be much easier for us to expand and implement those features as they come with a more separated endpoint schema. Then when the API begins reaching a stable state, we can mix and match those endpoints into larger endpoints that accept many, many arguments.

I'm also worried about how having few endpoints with many, many arguments would be tough to maintain and alter in the early days of this API. Like you've outlined a few times in our meetings, the goal is to move fast on this API and not be afraid to change or outright break things and I think having many endpoints with a clear focus will give us a lot of freedom in that process. (e.g. we can decided endpoint `x/y/z` is totally gone in favor of endpoint `a/b/c`)

manthey:
@banesullivan Any objection to merging this branch?  I'd like to add a view to let a user do searches, but I'd rather do it in a follow on PR with the changes I just did in #95).

Issue #88, opened 2020-07-07
Basic Support for Annotated Image Formats


banesullivan:
**Task:** Add support for annotated image formats [KWCOCO](https://gitlab.kitware.com/computer-vision/kwcoco) and JOIA (JOIA is not public)

I've decided to group these two formats together as their general structures are very similar and they will both require us to add a few new metadata fields and models to the `geodata` app. Also, we will not be supporting either format in full, so it makes sense to me to bundle them based on their overlapping concepts. By "not supported in full", I mean simplifying the fields we extract from either format and abstracting similar concepts across the two into our models as our goal is primarily to be able to ingest these data rather than to be able to output any imagery in our database to those formats.

Initial work is done in #91 

Implementing support for these formats will require:

- [x] Creating a new `Annotation` model that has a relationship to the `RasterEntry` model
    - [x] this will have a name/caption text field
    - [x] this will have a bounding box polygon
    - [x] a feature GeoJSON for label shapes (points/polygons/etc)
    - [x] this will have an optional "annotator" field that lists the name/entity of what made the annotation - this could be an algorithm, so this will just be a text field
    - [x] A general text field for notes
    - [x] JOIA and KWCOCO have "labels" - I think it makes sense just to add this as an optional text field.
    - [ ] KWCOCO annotations are defined in the image pixel space, so we will have to remap those annotations to world coordinates. - this will be dealt with when addressing #98 
- [x] Implementing a few more fields on the `RasterEntry` and `BandMeta` models:
    - [x] Add a checksum field for raster files - JOIA provides this, but should we compute a checksum on rasters by default?
    - [x] Add optional polygon field to `RasterEntry` for outlining non-null pixels
    - ~Perhaps add an optional version number?~
- [x] Possibly add a new `DerivedRaster` model that points to a base `RasterEntry` parent
    - JOIA has this concept of derived images that point to a parent image. This is useful in ML workflows where a processing routine alters the original image then annotates that image. I'm not sure what would be the best way to implement this model or if we should. Perhaps we just use the standard `RasterEntry` model for this and not track the parent image?
- [ ] Add basic proof of concept import/export routines for both formats.
    - waiting on #98 before starting this work
    - import KWCOCO with image annotations
    - wait for JOIA devs to decouple code from DB before making import script
    - add basic export routines for image data only (no support for annotations yet) for both formats

Some notes:

[...]

PR #91, opened 2020-07-09
Annotation Model, Checksum Files, & Cleanup

### LINKED ISSUE #97: Consistency across geodata relationships
the `RasterEntry` contains a reference to a `RasterFile`, but `GeometryArchive` contains a reference to a `GeometryEntry`.  

Let's change `GeometryEntry` to reference a `GeometryArchive` (since, at least, that way deleting a GeoemtryArchive will delete the associated GeometryEntry) for consistency across the geodata model relationships

I will try to include this in #91

banesullivan:
Add support for image annotations. Work in progress for #88 

I'd like to implement the ETL routines in another PR as KWCOCO might require us to tackle #98 first.

Todo:

- [x] Add routine to compute checksum and trigger a recompute of the checksum
- [x] Catchup with changes in #94 and make sure to propagate the refactoring that was done here to those changes
- [x] Resolve #97 

Notes:

- `ChecksumFile` is an abstract model that expects it's sub-table to have a `file` field. Should we make this non-abstract so that the checksum-ing can be run as a task? Would it make sense to put the `file` field in this model?... but how would we account for things like adding validators or specific upload locations? See [comment](https://github.com/ResonantGeoData/ResonantGeoData/pull/91#discussion_r452887767)
- I was going to do a few more things for the annotation, however, I am thinking about what will be necessary for #98 and we will likely have to completely redesign the `Annotation` model for that so I'd rather wait to improve that model until after #98  is resolved.


rgd/utility.py
```diff
@@ -23,3 +24,11 @@ def _field_file_to_local_path(field_file: FieldFile) -> Generator[Path, None, No
                 yield Path(dest_stream.name)
         else:
             yield Path(file_obj.name)
+
+
+def compute_checksum(file_path, chunk_num_blocks=128):
+    sha256 = hashlib.sha256()
```
manthey, line 30: In Girder, we often use sha512.  Perhaps we should here for consistency?

banesullivan: I have no opinion/experience on what to use. I chose sha256 over md5 because:

> Unless otherwise noted, sha256 is the RECOMMENDED algorithm for checksums in
JOIA.

From https://gitlab.kitware.com/vigilant/joia/-/blob/master/joia/docs/schema/Overview.md#conventions

manthey: Then sha256 makes sense.

geodata/models/common.py
```diff
@@ -50,6 +43,58 @@ def __str__(self):
...
+            self.validate_checksum = False
+            # Simple update save - not full save
+            super(ChecksumFile, self).save(
+                update_fields=[
+                    'checksum',
+                    'compute_checksum',
+                    'last_validation',
+                    'validate_checksum',
+                ]
+            )
```
banesullivan, line 94: @manthey, do we want to run this checksum-ing as a task? I tried doing so, but I couldn't think of a clever way of accessing any given subclass model. Since I made this an abstract class and expect child classes to have a `file` field, there's no manager for this model and I'm not sure how to easily query for this model entry within the task.

PR #104, opened 2020-07-16
Generate serializers dynamically

mattrix27:


manthey:
We are going to want the ability to specify some serializers explicitly.  Ideally, we'd only create serializers for models that don't already have them.  This could probably be something like:
```
for _model_name, model in inspect.getmembers(models):
      if not inspect.isclass(model):
          continue
      parent = model
      while len(parent.__bases__):
          if base_models.Model in parent.__bases__:
              break
          parent = parent.__bases__[0]
      if base_models.Model in parent.__bases__:
          model_fields = {}
          serializer_name, serializer_class = create_serializer(model, model_fields)
          if serializer_name not in globals():
              globals()[serializer_name] = serializer_class
```
in `geodata/serializers.py` and `core/serializers.py`.  This could be placed in a utility function, where those serializer files would call `make_serializers(globals(), models)` or something similar.

geodata/search.py
```diff
@@ -90,7 +90,8 @@ def search_near_point_filter(params):
 def search_near_point(request, *args, **kwargs):
     params = request.query_params
     results = SpatialEntry.objects.filter(search_near_point_filter(params))
-    return JsonResponse(serializers.SpatialEntrySerializer(results, many=True).data, safe=False)
+    serializer = utility.create_serializer(SpatialEntry)
```
manthey, line 93: We probably don't want to create a class each time this function is called.  If the serializers are generated once (on import, for instance), then they could still be referenced by name as before.

rgd/utility.py
```diff
@@ -42,3 +49,97 @@ def _link_url(root, name, obj, field):
...
+            'queryset': model.objects.all(),
+            'serializer_class': serializer,
+            'filter_backends': [DjangoFilterBackend],
+            'filterset_fields': get_filter_fields(model),
+        },
+    )
+    return viewset_class
+
+
+def make_serializers(globe, models):
```
manthey, line 131: Can we rename `globe` to something a little more indicative of its purpose, perhaps `serializer_scope`.  This should probably also get a docstring describing its purpose, as it is definitely cryptic.  Something like 
```
"""Make serializers for any model that doesn't already have one.

This should be called after specific serializer classes are created.  Serializers are created named <model_name>Serializer.

:param serializer_scope: the scope where serializers on the models are defined.  In a serializers.py file, this will be globals().
:param models: a namespace with defined models for which to create serializers.  This can be a models module.
"""
```

manthey:
I merged a branch to master that will cause minor conflicts with this one.  Please merge master and resolve those conflicts.  Otherwise, with the variable name change and added comment, this looks good.

Issue #82, opened 2020-06-24
Tests for the ETL routine


banesullivan:
We should write tests to verify the ETL routine (the readers under `geodata/models/`) work on CircleCI

aashish24:
Daniel C is doing automated testing for the DANDI project as per Brian so perhaps it is worth talking to him

banesullivan:
note to self: see work done in #107 on how to get started. Brian did an awesome job of setting up the testing infrastructure -- now it's up to me to start filling in the tests for the `geodata` app.

banesullivan:
We will inevitably need to store some rasters and geometry for these tests. @brianhelba and @manthey, where might be the best place to store these data? Should we include these in the repository directly or maybe create another repo under @ResonantGeoData that we can clone during the tests?

manthey:
I like what I've done in large_image and histomicsui.  Specifically, I have a python utility function that gets used in pytest functions.  It could be made into a fixture, if that were desired.  External data is hosted on data.kitware.com (though it could be hosted anywhere that is publicly accessible or accessible with stored credentials).

In the main repo, we store a key file which contains the sha512 hash of each data file we want from the external data store.  For instance, in large_image those sha512 files are in a test/data directory (https://github.com/girder/large_image/tree/master/test/data).  In the pytest function, these get used like `filePath = utilities.externaldata('data/sample.ome.tif.sha512')`.  This file is fetched and stored in a subdirectory of the tox directory.  If you run tox a second time, the hash of the file is checked and the file is only downloaded again if it has changed.

The utility function just uses `requests`.  In principal, it could use any service that can download a file based on a hash value.

This allows stored large test files, since they aren't in the repo.  GitHub limits file size to 100Mb.  There is git-lfs, but that costs per download after some threshold.

brianhelba:
I think Git LFS is the best option. Do you think we'll exceed [GitHub's limits](https://docs.github.com/en/github/managing-large-files/about-storage-and-bandwidth-usage) in our initial usage? I'd hope that if only for test performance, we'd keep the sample raster files as small as possible.

To be fair, Git LFS has a decent number of flaws and annoyances, but I think it's the most generally supported and most resilient option.

brianhelba:
@banesullivan I can set up the Git LFS infrastructure in this repo, such that we have a dedicated directory to commit test files. We should just resolve the question about expected size before committing to that route.

manthey:
[...]

PR #105, opened 2020-07-16
Fix outline and create footprint of valid raster data

### LINKED ISSUE #100: Improve the footprint of RasterEntry
#91 adds an `outline` field for the total footprint of rasters. For the moment, the reader sets the inherited `footprint` field to that outline.

@manthey suggests: 

> the `footprint` we query by default should have the boundary of non-null pixels, and, if we want the raster coverage including null pixels that should be the alternately named entry.  In general, I would expect someone searching for data to only care if data is present.  For some formats (WAMI), there can be a large area with functionally null pixels (though, confusingly, the WAMI data has both nodata values and background values that should have been marked nodata but instead were stored as white).

I think that is a very valid point. In most search applications that I have seen, the full footprint is returned and it is often very misleading.

@manthey:

> One idea is to have a field with the "total_footprint" or some other name with what we have in the footprint now; the footprint would be this if we haven't computed a data mask, or the data mask if we have.  Ideally we'd add information about when and how the data mask was extracted.  For instance, on WAMI data, I wrote a simple convex hull routine to find the data mask, but later wrote a (unfinished) routine to better fit the data as it had concavities.  We'd mostly query the data mask as stored in the footprint.

----

In summary, we need:

- [x] Continue to store the total footprint of the raster in `outline` and add it to `footprint` by default (this is what #91 does)
- [x] Add a way to compute a polygon of the convex hull of non-null data for all rasters and make sure this works across bands
- [x] Store that data mask in the `footprint` field updating the default `outline`

### LINKED ISSUE #103: Footprints may not be correct
A spatial reference is not completely captured by the srid, resulting in bad footprint polygons.

I have a sample file where the coordinate system is wkt 

```
PROJCS["Albers Conical Equal Area",GEOGCS["NAD83",DATUM["North_American_Datum_1983",SPHEROID["GRS 1980",6378137,298.2572221010042,AUTHORITY["EPSG","7019"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY["EPSG","6269"]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433],AUTHORITY["EPSG","4269"]],PROJECTION["Albers_Conic_Equal_Area"],PARAMETER["standard_parallel_1",29.5],PARAMETER["standard_parallel_2",45.5],PARAMETER["latitude_of_center",23],PARAMETER["longitude_of_center",-96],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]]]
``` 

When we create the raster footprint, we have a `SpatialReference` with this wkt and we create a `Polygon` with `srid=spatial_ref.srid`, but the `srid` is just `4269`, which is not in the same units as the actual spatial reference.  Specifically, we are trying to do: 

```
Polygon([[1319547.040429464, 2658548.125730889], [2005547.040429464, 2658548.125730889], [2005547.040429464, 2148548.125730889], [1319547.040429464, 2148548.125730889], [1319547.040429464, 2658548.125730889]], srid=4269)
```

but this doesn't work.

A sample file:
[landcover_sample_2000.zip](https://github.com/ResonantGeoData/ResonantGeoData/files/4927886/landcover_sample_2000.zip)

I think we need to convert our boundary polygon before creating a Django GEOSGeometry polygon.  Another option might be to register each coordinate system we use via `django.contrib.gis.utils.srs.add_srs_entry`, but I haven't checked if that would work.

banesullivan:
Potentially resolve #103 and resolve #100 

Also, this adds a routine to compute the convex hull of the valid data points within the raster. @manthey, what do you think of this? Is there a simpler way to do this? 

Also, is it okay to add scipy as a depend?

I managed to get this to work with the transform approach and now the footprint is showing in the correct location on the geodjango map:

Screen Shot 2020-07-16 at 3 36 10 PM


geodata/models/geometry/transform.py
```diff
@@ -0,0 +1,15 @@
...
+from osgeo import ogr
+from osgeo.osr import CoordinateTransformation, SpatialReference as SpatRef2
+
+from ..constants import DB_SRID
+
+
+def transform_geometry(geometry, source_wkt):
+    """Transform geometry into the database's spatial reference system."""
+    g = ogr.Geometry(wkt=geometry.wkt)
+    source = SpatRef2(source_wkt)
```
manthey, line 12: One of the major issues between Django and GDAL >= 3 is that Django always expects axes to be x-y, but GDAL 3 defaults to the CRS authority order.  Django 3.1 explicitly does `<spatial_reference>.SetAxisMappingStrategy(osgeo.osr.OAMS_TRADITIONAL_GIS_ORDER)`.  We probably need to do that on `source` and `dest`, as, otherwise, the exported geometry may have its coordinates in y-x order.

I have some test cases that should show if this is necessary.

Since this does the geometry transform entirely in GDAL, and only exchanges it with Django via wkt, it looks like it should work generally.

banesullivan: > Since this does the geometry transform entirely in GDAL, and only exchanges it with Django via wkt, it looks like it should work generally.

That's what I was hoping. By keeping this transform entirely managed by GDAL, hopefully, this means we can send any geometry to this method and it will output it in the DB spatial reference across version combinations of Django and GDAL

> I have some test cases that should show if this is necessary.

I've been testing this with Landsat imagery (which has the Y,X convention I think) and all seems to be working. It'd be good to get a few other test datasets that we can host somewhere for the tests that we really need to get to in #82

banesullivan: > .SetAxisMappingStrategy(osgeo.osr.OAMS_TRADITIONAL_GIS_ORDER)

I had initially tried that yesterday and it yielded the wrong results with my version of Django and GDAL: `3.0.7` and `3.1.0` respectively

geodata/models/raster/reader.py
```diff
@@ -2,17 +2,55 @@
...
+    src is an open dataset with rasterio
+
+    Returns a numpy array of the bounadry points in a closed polygon.
+
+    """
+    msk = src.read_masks(band_num)
+
+    a = (np.arange(msk.shape[1]) * src.res[1]) + src.bounds.left
+    b = (np.arange(msk.shape[0]) * src.res[0]) + src.bounds.bottom
+    xx, yy = np.meshgrid(a, b[::-1])
```
banesullivan, line 44: These three lines also make an assumption about the coordinate system. We may want to wait until after the transform to do this

manthey:
I tried this on a Lair WAMI image and got the following error:
```
[2020-07-17 14:28:17,041: INFO/ForkPoolWorker-2] geodata.tasks.validate_raster[8dc64ab7-6cdc-479c-a46b-d8c51d2e39d1]: Issue computing convex hull of non-null data: QH6235 qhull error (qh_memalloc): negative request size (-2053897440).  Did int overflow due to high-D?

While executing:  | qhull i Qt
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 60381305  incidence  Qtriangulate  _pre-merge  _zero-centrum
  _max-width 0.068  Error-roundoff 7.2e-14  _one-merge 3.6e-13
  _near-inside 1.8e-12  Visible-distance 1.4e-13  U-max-coplanar 1.4e-13
  Width-outside 2.9e-13  _wide-facet 8.6e-13  _maxoutside 4.3e-13

double free or corruption (out)
```
Based on that malloc size, it looks like it tried to malloc w*h*4+200 bytes (the image is 26368 x 21248)

banesullivan:
> Based on that malloc size, it looks like it tried to malloc w * h * 4+200 bytes (the image is 26368 x 21248)

🤣 welp. Looks like we need a better/more efficient way to compute the convex hull of valid data. @manthey, have you ever done this in the past with large rasters?

geodata/models/raster/reader.py
manthey: Also, this is using the corner of a pixel as representative of the point.  Pedantically, this means right and top edges will be excluded from the returned convex hull.

banesullivan:
I could downsample the `mask` array that `rasterio` loads via its `out_shape` argument: https://rasterio.readthedocs.io/en/latest/api/rasterio._io.html?highlight=out_shape#rasterio._io.DatasetReaderBase.read_masks

manthey:
Yes.  There are two main approaches: downsampling or tiling.

With downsampling, the danger is that with nearest neighbor you can miss small features if using nearest neighbor.  I've done it where the convex hull is guaranteed to have all of the data, but might have more area because the resampling was done via any weighted generation technique.  One virtue of downsampling is that it is much faster than tiling.

With tiling (or striping), you can generate a convex hull for each tile and then combine them when done.  Or, you can find the extremes of data on each raster line and generate a convex hull from those extrema (really just getting the convex hull per line and then combining them, but often faster than using actual convex hull algorithms).

manthey:
I think I would choose to downsample using averaging or bilinear (since we don't really care what the downsampled value is), and accept that the computed convex hull may be slightly larger than the actual data.

geodata/models/raster/reader.py
banesullivan: @manthey, I've updated this to compute the coordinates at all cell centers. This means there are XY coordinates for every cell in the mask array rather than on the lower-left edges of all cells

banesullivan: Actaully, I messed this up. Will fix again

banesullivan: Okay, now it is fixed

banesullivan:
I went with the downsampling approach setting a maximum resolution of 4K by 4K.

The resampling is handled by rasterio and it is set to use an averaging algorithm. We can also use bilinear if there is a preference.

PR #106, opened 2020-07-20
expose geodata models

mattrix27:
Definitely needs review

I had to exclude a lot of fields for swagger to generate because the default filter that we currently use doesn't support models from `contrib.gis` and `contrib.postgres`. Will probably have to find specific filter_backends or create custom filters or something

Also trying to make it part of its own endpoint (like 'api/geodata'), but using router makes it complain


manthey:
I think with `django-rest-framework-gis`, if we add it to the installed apps, then it will automatically serialize all geometry fields to geojson.

Also, if we place geodata in `api/geodata`, we should also move core to `api/core`.  Based on current project specifications, geodata is actually the main API and what we've called core is secondary (which means we might want to rename core to something else more indicative of what it is -- but we'll discuss that elsewhere).

rgd/utility.py
```diff
@@ -64,6 +69,11 @@ def parse(self, stream, media_type=None, parser_context=None):
             if isinstance(getattr(model, key), fields.related_descriptors.ManyToManyDescriptor):
                 for val in value.split(','):
                     qdict.update({key: val.strip('"')})
+            elif model.__name__ == 'RasterEntry' and key in raster_arrays:
```
manthey, line 72: We really don't want to leak our models into the general utility function.  It will cause future maintenance problems.  If django-rest-framework-gis is added to the installed apps, is this still necessary?

geodata/urls.py
```diff
@@ -19,4 +26,5 @@
     path('api/geodata/near_point/extent', search.search_near_point_extent),
     path('api/geodata/raser/near_point/extent', search.search_near_point_extent_raster),
     path('api/geodata/geometry/near_point/extent', search.search_near_point_extent_geometry),
+    path('api/geodata/', include(router.urls)),
```
manthey, line 29: I'm seeing this mounted at `/api/geodata/api/<model name>`.  Because swagger automatically hides common prefixes, it shows up in swaggeras `geodata/api/<model name>`.  If we mount everything under `geodata`, the swagger interface is grouped poorly.  As a consequence, I think we just want to mount this via `path('', include(router.urls))` and let the API list both the core and geodata models.  If we have models with the same names in each app, then my preference would be to override core, since it is actually our less important use case.

mattrix27: Just to clarify, Dataset is in both, right now it overrides geodata's, so should I make it so that core's get overridden or should I just rename one of the models?

manthey: For now, don't worry about which one wins (since I assume it has to do with import order).  Just make a new issue to revisit how to resolve this once the rest of this PR is done.

rgd/utility.py
mattrix27: So overall, all of the geometry fields are already being serialized to geojson from what I can tell looking at my sample queries, but these raster_array fields are from `django.contrib.postgres`  so installing django-rest-framework-gis didn't seem to help.

mattrix27: Did a simple fix to not leak models

manthey:
The listing and read paths work as expected.  If I expand a POST or PATCH endpoint (and probably a PUT endpoint, too), I get some javascipt exceptions that break swagger.  One such error reads `Could not find component JsonSchema_string_date-time` , though there are other components it doesn't find, too.

mattrix27:
I created serializers for RasterEntry and Geometry so now it doesn't ask for a spatialentry_ptr when POSTing. It will automatically create a SpatialEntry with the footline and outline when POSTed. If we don't want this or want to get rid of even more then we just need to edit the read_only_fields in the serializers. Depending on what we want the API to do we might want to just add something to exclude the model altogether, but that shouldn't be too difficult

PR #107, opened 2020-07-21
Use improved settings and Docker config from upstream

brianhelba:
This significantly reduces the amount of downstream infrastructure code in RGD, while providing several requested development features, such as the ability to remap service container ports.

This should resolve the issue where a full "Develop with Docker" configuration is unable to download files from MinIO. A resolution to the issue of being unable to cleanly upload files directly to MinIO via S3FF is currently under development.


brianhelba:
I've also pushed 2 more commits which update and fix the pytest testing stack, and adds some example test stubs to the geodata app.

docker-compose.yml
```diff
@@ -1,25 +1,30 @@
...
     volumes:
-      - "dbdata:/var/lib/postgresql/data"
+      - dbdata:/var/lib/postgresql/data
 
   rabbitmq:
     image: rabbitmq:management
+    ports:
+      - ${DOCKER_RABBITMQ_PORT-5672}:5672
+    volumes:
+    - mqdata:/var/lib/rabbitmq
```
brianhelba, line 18: Remove this, we never want to persist the state of a RabbitMQ volume across container re-creation.

PR #108, opened 2020-07-22
Set the basepython for tox.

manthey:
Tox generally expects tests to be run with names like "py38" not "test".  If that is done, then the basepython is automatically extracted from the environment.  Specifying the basepython allows tox to run successfully in a multi-python environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,6 +3,9 @@ envlist =
     lint,
     test,
 
+[testenv]
+basepython=python3.8
+
 [testenv:lint]
 skipsdist = true
 skip_install = true


brianhelba:
This will prevent Tox from being used with Python 3.9 in the future. When we do want to support running under any version other than exactly 3.8, this will need to be updated.

manthey:
True, which is why it is more common to use environment names like `py38` than `test`.

brianhelba:
I think it's more complicated than this. Several other projects were putting the pytest execution under the top-level `[testenv]` section. Normally, this can be harmless, as all of the sections (`deps`, `commands`) are typically overridden by other environments. However, if you add anything beyond this to the top-level section to facilitate pytest, the other environments inherit it and potentially break.

If we want to explicitly support running Tox against multiple Python versions, we can add [an environment specifier](https://tox.readthedocs.io/en/latest/config.html#tox-environments), so you'd run the environment `test-py38`.

manthey:
Here, we want to make sure `lint`, `format`, and `test` are all run in Python 3.8.  Otherwise, if you are using this in an environment where the default Python isn't 3.8, tox will use that Python (linting will fail on Python 3.7, for instance, since we use 3.8 features).  If there is a way to specify a *minimum* version of Python, that would be better, but I don't see how to do that in the tox documentation.

manthey:
My goal is to have `tox` work by default.  It seems poor to require `tox -e lint-py38,test-py38` as a default command.

PR #109, opened 2020-07-22
Redesign GeoData for non-spatial models

### LINKED ISSUE #98: Redesign geodata models to support non-spatial data
Per team discussion, we need to support non-spatial imagery as well.

I will brainstorm a new model hierarchy and post it here, then sync up with David about the implementation. I already have a few ideas about how we can make this work in a general fashion while keeping the `SpatialEntry` concept that informed our initial model design.

banesullivan:
Resolve #98 - see #98 for UML diagram, details, and design discussion. 

This PR will wipe out all migrations for the GeoData app and start fresh. Should we do the same for the `core` app while we're at it?

Todo:

- [x] Make sure task launching with pre/post save events works okay. It may be in an endless cycle of saving, launching a task, saving within the task, then launching a new task from that save

Current models (will update as PR evolves):

geodata


banesullivan:
This is caught up with the changes in #115

geodata/models/__init__.py
```diff
@@ -3,4 +3,4 @@
 from .common import SpatialEntry  # noqa
 from .dataset import Dataset  # noqa
 from .geometry.base import GeometryArchive, GeometryEntry  # noqa
-from .raster.base import RasterFile, RasterEntry, ConvertedRasterFile  # noqa
+from .imagery.base import ImageFile, RasterEntry, ConvertedImageFile  # noqa
```
manthey, line 6: We should add `ImageSet` to this so it is exposed at the top of the `geodata.models` namespace.

geodata/models/imagery/base.py
```diff
@@ -0,0 +1,96 @@
...
+class ImageSet(ModifiableEntry):
+    """Container for many images."""
+
+    name = models.CharField(max_length=100, blank=True, null=True)
+    description = models.TextField(null=True, blank=True)
+
+    images = models.ManyToManyField(ImageEntry)
+
+
+class RasterEntry(SpatialEntry, ImageSet, TaskEventMixin):
```
manthey, line 48: If we change the order of inheritance to `ImageSet, SpatialEntry, TaskEventMixin`, then name becomes the first column.  Currently, acquisition date is the first column.  Alternately, we could change the admin view to have name first.

banesullivan: Ahh, I did not realize that happened. Good to learn how inheritance order affects the primary key/first column

geodata/admin.py
```diff
@@ -15,61 +20,74 @@
...
         '__str__',
         'modified',
     )
     readonly_fields = (
-        'number_of_bands',
-        # NOTE: we do not want users to edit these, but the geodjango interface is really nice
-        # 'footprint',
-        # 'outline',
-        'raster_file',
+        'footprint',
```
manthey, line 57: We can still show a map.  Set `modifiable = False` as a class property and the maps are read-only, then remove the `footprint` and `outline` fields from the `readonly_fields` dictionary.

banesullivan: Done. The only caveat that I see to this is that when creating a RasterEntry, the map is shown for the user to select these polygons. I had to set `blank=True` to those fields to allow the user to not fill those out since they are auto-populated

banesullivan:
Agreed. The only item left on this is to have `read_geometry_archive` set the `outline` field which it currently leaves blank. For some reason, I still have GDAL issues with where the site is running locally for me, so I cannot test any geometry ETL routines.

The outline should be the bounding box of the `footprint`: `[[xmin, ymin], [xmin, ymax], [xmax, ymax], [xmax, ymin] [xmin, ymin]]`

banesullivan:
This is almost ready to merge. For the moment, we cannot delete any `SpatialEntry` models from the django admin interface. I get the error `'SpatialEntry' object has no attribute 'id'` which is odd because I set the primary key of `SpatialEntry` to `spatial_id` so the admin interface should know to use that?

docker-compose.override.local-mac.yml
```diff
@@ -0,0 +1,6 @@
+version: '3'
+services:
+  celery:
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - /usr/local/bin/docker:/usr/bin/docker
```
manthey, line 6: Would it be better to just change `docker-compose.override.yml` to have
```
 volumes:
      - ${DOCKER_SOCK:-/var/run/docker.sock}:/var/run/docker.sock
      - ${DOCKER_CMD:-/usr/bin/docker}:/usr/bin/docker
```
and then you would override this with environment variables?

geodata/models/common.py
```diff
@@ -29,17 +26,21 @@ def save(self, *args, **kwargs):
...
-class SpatialEntry(ModifiableEntry):
-    """Common model to all geospatial data entries."""
+class SpatialEntry(models.Model):
+    """Common model to all geospatial data entries.
+
+    This is intended to be used in a mixin manner.
+
+    """
+
+    spatial_id = models.AutoField(primary_key=True)
```
manthey, line 36: You'll need to change `def __str__(self):` to match, e.g., 
```
    def __str__(self):
        return '{} (type: {})'.format(self.spatial_id, type(self))
```

Or just remove the custom string method.

docker-compose.override.local-mac.yml
banesullivan: Much simpler

banesullivan:
@manthey, all should be good at this point. Ready to merge when you are

manthey:
@banesullivan Merge away so we can start adding tests, etc.

PR #115, opened 2020-07-27
Implement an initial simple raster search and display.

manthey:
This also fixes some coordinate transform issues.  This requires upgrading to django 3.1 (which is currently in rc1, and is not
official).  Unfortunately, there is definitely code in djangorestframework with will fail in django 3.1 until drf has a new release.

Specifically, there is ambiguity in how django 3.0 handles GDAL 3 transforms.  Compounding the problem, rasterio is bundled with a GDAL 2.x library.  Django should be using our more complete GDAL 3 library, but depending on the import order of rasterio and osgeo, rasterio may not emit wkt coordinates in a consistent manner.

### DIFF
--- a/geodata/models/raster/reader.py
+++ b/geodata/models/raster/reader.py
@@ -109,6 +109,14 @@ class RasterEntryReader(_ReaderRoutine):
             self.raster_entry.creator = self.rfe.creator
             self.raster_entry.modifier = self.rfe.modifier
 
+            # There is a potential conflict between rasterio and whatever GDAL
+            # is available.  Rastio has an older form of GDAL and conflicts
+            # with a system GDAL if the version is different.  So far, the only
+            # issue seems to be with rastio's <source>.crs.  We can work around
+            # this by using GDAL directly.
+            gsrc = gdal.Open(str(file_path))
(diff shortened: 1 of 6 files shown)


manthey:
I'm marking this as a draft until I've added tests.  The tests I intend to add first are for the ETL to ensure that the coordinates that are retrieved for the data sets are the expected values.

geodata/models/raster/reader.py
```diff
@@ -109,6 +109,14 @@ def _read_files(self):
...
             self.raster_entry.modifier = self.rfe.modifier
 
+            # There is a potential conflict between rasterio and whatever GDAL
+            # is available.  Rastio has an older form of GDAL and conflicts
+            # with a system GDAL if the version is different.  So far, the only
+            # issue seems to be with rastio's <source>.crs.  We can work around
+            # this by using GDAL directly.
+            gsrc = gdal.Open(str(file_path))
+            spatial_ref_wkt = gsrc.GetSpatialRef().ExportToWkt()
+            spatial_ref = SpatialReference(spatial_ref_wkt)
```
banesullivan, line 119: @manthey, this change will conflict with my work, but I can easily resolve it in #109

PR #117, opened 2020-07-31
Add some aggregation properties to raster entries.

manthey:
There might be a better way to do this, but it avoids storing redundant data in the database.


manthey:
@banesullivan If you approve this, just merge it into your non-spatial branch.

image

PR #118, opened 2020-07-31
Improve swagger descriptions.

manthey:
Some swagger clients show summaries or descriptions rather than the endpoint as the primary menu item.  As such, make sure these are populated.

We should always populate the `operation_summary` if we specify an `operation_description` in `swagger_auto_schema`.

If we add model __doc__ strings, when we auto-generate views to generate swagger API, the first line become the summary and the rest of the docstring becomes the description.

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -25,6 +25,13 @@ class DeferredFieldsManager(models.Manager):
 
 
 class Task(models.Model):
+    """
+    Data processing task.
+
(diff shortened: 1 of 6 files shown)


manthey:
Before:
image
After:
image

manthey:
This makes a bigger difference on other swagger UIs:
Before:
image
After:
image

PR #120, opened 2020-08-04
Autopopulate ImageSet names.

manthey:
This also shows ImageEntry names as part of their __str__ to make it easier to select them when making an ImageSet or RasterEntry.

### DIFF
--- a/geodata/models/imagery/base.py
+++ b/geodata/models/imagery/base.py
@@ -44,6 +44,9 @@ class ImageEntry(ModifiableEntry):
 class ImageSet(ModifiableEntry):
     """Container for many images."""
 
+    def __str__(self):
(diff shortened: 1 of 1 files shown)


geodata/models/imagery/base.py
```diff
@@ -69,6 +85,9 @@ class RasterEntry(ImageSet, SpatialEntry, TaskEventMixin):
 
     """
 
+    def __str__(self):
+        return f'{self.name} ({self.id} - {type(self)}'
+
```
banesullivan, line 90: Perhaps we should move this to `ImageSet` so it also has this functionality and `RasterEntry` will inherit it

manthey: Will do.

PR #121, opened 2020-08-04
Add test for rasterfile centroids.

manthey:


manthey:
When creating many-to-many entries in a test factory, somehow the post commit hooks are processed before the factory saves  the model.  The factory resaves the model (internally in factory_boy after processing), but, since this hasn't refreshed the model from the post commit hook changes, obliterates those changes.  Currently, I've worked around this issue in the RasterEntryFactory by manually running the post commit hook.  We'll have to dig deeper into factory_boy to figure out how to avoid needing to do this.

I think we should merge this as it stands.

manthey:
@banesullivan This now changes some of the TaskEventMixin code.  It makes the test factory less hacky, though, and it will work in a synchronous environment (whereas, before, it would not). 

I think this is ready for review and merging.

PR #123, opened 2020-08-10
Coverage Reporting

banesullivan:
Add coverage reporting to codecov.io and generate HTML coverage reports locally

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -85,3 +85,6 @@ ignore =
 DJANGO_SETTINGS_MODULE = rgd.settings
 DJANGO_CONFIGURATION = TestingConfiguration
 addopts = --strict-markers --showlocals --verbose
+
+[coverage:run]
+branch = True
--- a/tox.ini
+++ b/tox.ini
@@ -52,8 +52,10 @@ deps =
     pytest-django
     pytest-factoryboy
     pytest-mock
+    pytest-cov
 commands =
-    pytest {posargs}
+    pytest --cov=rgd --cov=core --cov=geodata {posargs}
+    coverage html
 
 [flake8]
 max-line-length = 2048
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -52,8 +52,10 @@ deps =
     pytest-django
     pytest-factoryboy
     pytest-mock
+    pytest-cov
 commands =
-    pytest {posargs}
+    pytest --cov=rgd --cov=core --cov=geodata {posargs}
+    coverage html
```
banesullivan, line 58: I added this html report generation so I could easily see what's hitting/missing during development. If this is too much overhead for our testing script, I can let it be something I only do locally when developing

banesullivan:
Coverage is now reporting on codecov.io! See [coverage for this branch](https://codecov.io/gh/ResonantGeoData/ResonantGeoData/branch/coverage)

tox.ini
manthey: Tools work for us.  If you find the html useful, then we should generate it.

manthey:
This is great.  I would recommend changing where the html coverage is emitted to avoid putting it directly in the project directory.  That is, I'd change
```
coverage html
```
to
```
coverage html -d {toxworkdir}/htmlcov
```
so it ends up at `.tox/htmlcov`.

Additionally, I use a tool that uses the xml for showing coverage.  It would be handy for my existing workflow to also add
```
coverage xml -o {toxworkdir}/coverage.xml
```

banesullivan:
I updated the tox.ini to place the coverage reports (HTML and XML) in the `.tox` directory

manthey:
Looking at the reports, let's also turn on branch coverage reporting.  This can be done by adding the appropriate flag in the tox.ini file:
```
[coverage:run]
branch = True
```

PR #126, opened 2020-08-11
Require a working version of S3FF to force upgrades

brianhelba:
This fixes the issues with Docker Compose Minio not respecting uploads. In particular, S3FF now respects the `MINIO_STORAGE_MEDIA_URL` setting.

@manthey Is there any workaround code which should be removed?


brianhelba:
Closing in favor of #132

PR #129, opened 2020-08-11
Add development extras

brianhelba:


setup.py
```diff
@@ -44,4 +44,5 @@
         'shapely',
         'scipy',
     ],
+    extras_require={'dev': ['ipython', 'tox']},
```
manthey, line 47: What are you using ipython for?

banesullivan: It's nice to have for the `./manage.py shell` interface when debugging

brianhelba: Exactly. Particularly for `./manage.py shell_plus`.

I'd like to add some docs on these development tools and techniques at some point.

PR #132, opened 2020-08-13
Handle changes to s3-file-field.

manthey:
Don't import gis functions in utility module; it can get imported before the settings have found GDAL.


manthey:
@brianhelba This is a superset of #126, since the newest version of S3FF renamed its configurations module.

I haven't yet checked if we can get rid of workaround code.  I'd rather get this merged it since otherwise the project is broken and then deal with code cleanup on another PR (possibly on yours).

PR #134, opened 2020-08-14
KWCOCO import support

### LINKED ISSUE #88: Basic Support for Annotated Image Formats

banesullivan:
Resolve #88 - be sure to squash and merge!

Note that this resets all the migrations in the geodata app

Todo:

- [x] Ensure RLE masks work
- [x] Add tests
- [x] validate against real data
- [x] Clean up migrations into a single file
- [x] If `KWCOCOArchive` is deleted, make sure to delete the associated `ImageSet`
- [x] Do we want an `ImageFile` entry for each image in the KWCOCO dataset. At present, there is no way to track back where an image came from if imported in KWCOCO format. We'll just have the metadata and annotations for the image if we wanted to export one specific image and it's annotations, it'd be tough to find the original image.


setup.py
```diff
@@ -43,6 +43,8 @@
         'fiona',
         'shapely',
         'scipy',
+        'kwcoco',
+        'kwimage',
```
banesullivan, line 47: This adds: 

```
ordered-set, ubelt, pandas, torch, kwarray, opencv-python, Pillow, imageio, PyWavelets, cycler, kiwisolver, matplotlib, decorator, networkx, tifffile, scikit-image, kwimage, kwplot, bezier, parse, joblib, threadpoolctl, scikit-learn, pyyaml, scriptconfig, kwcoco
```

dev/django.Dockerfile
```diff
@@ -8,6 +8,8 @@ RUN apt-get update && \
         gcc \
         libc6-dev \
         libmagic1 \
+        libgl1-mesa-glx \
+        libglib2.0-0 \
```
banesullivan, line 12: Necessary for `kwimage`

geodata/migrations/0004_auto_20200903_1811.py
```diff
@@ -0,0 +1,68 @@
...
+            ),
+            preserve_default=False,
+        ),
+        migrations.AlterField(
+            model_name='imageentry',
+            name='image_file',
+            field=models.OneToOneField(
+                null=True, on_delete=django.db.models.deletion.CASCADE, to='geodata.baseimagefile'
+            ),
+        ),
```
banesullivan, line 38: This might not be right... Since I am adding an inherited model, this might bot be the right migration to do.

PR #141, opened 2020-09-02
Remove unnecessary Flake8 excludes

brianhelba:
The default exclude rule should be adequate for all files in the repo.

Also, fix a minor typo in the README

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,13 +63,6 @@ commands =
 [flake8]
 max-line-length = 2048
 show-source = True
-exclude =
-    env37
-    env38
-    .git
-    .tox
-    __pycache__
-    */*egg*/*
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -63,13 +63,6 @@ commands =
 [flake8]
 max-line-length = 2048
 show-source = True
-exclude =
-    env37
-    env38
-    .git
-    .tox
```
manthey, line 70: When I run `tox -e lint` locally, if `.tox` is not excluded, tox appears to hang because it takes so long.  Although it might eventually get over it, it seems like excluding `.tox` is necessary.

brianhelba: I'll look into this.

PR #143, opened 2020-09-07
FMV ETL Routine

### LINKED ISSUE #142: FMV Support
This issue is a place to track progress and details for implementing support for Full Motion Video (FMV)

FYI: [this is helpful](https://www.remotegeo.com/demystifying-misb-fmv-full-motion-video/) for those unfamiliar with FMV.

### Available Data

The VIRAT AP Hill data is the closest thing that does not come with any usage restrictions. It is publicly released and has KLV metadata. This dataset is hosted by Kitware [here](https://data.kitware.com/#collection/56f56db28d777f753209ba9f/folder/56f57fa48d777f753209c713). 

### Spatial Reference / Metadata

- The KLV metadata is encoded in a video file (e.g. MP4) and has geospatial metadata included in that file in the KLV format.
- Most other data sets either have no metadata, or the metadata is stored in a separate file (e.g. CSV), or the "video" is actually a sequence of JPEGs and the metadata is stored as EXIF in the JPEG headers.
- We should focus on KLV metadata as that's what our customers/users are interested in.
- KLV is in a separate stream within the video file, much like audio is.

We can use Kwiver to extract this FLV metadata with the `kwiver dump-klv` command-line tool.  Source [here](https://github.com/Kitware/kwiver/blob/master/arrows/core/applets/dump_klv.cxx). To use it, download the Docker image published for Kwiver [here](https://hub.docker.com/r/kitware/kwiver/tags) and run the tool with the video file. For example:

```
HOST_MNT=$PWD/
CONT_MNT=/home/
CONT_PWD=/home/

docker run --rm -v $HOST_MNT:$CONT_MNT:delegated -w $CONT_PWD kitware/kwiver:release /kwiver/build/bin/kwiver dump-klv VIRAT/09152008flight2tape1_1.mpg
```

where `09152008flight2tape1_1.mpg` saved in the `VIRAT` folder and is the [first video](https://data.kitware.com/#item/56f57fa48d777f753209c714) in the dataset.

Note that we are currently having trouble with that Docker image as Kwiver might not be properly configured with FFmpeg. See https://github.com/Kitware/kwiver/pull/1070 for details and a potential fix.

At the moment, I am roadblocked from continuing this until the Docker image is fixed and we can use it to extract the FLV metadata.

### Next Steps

- [x] Use an updated Kwiver Docker image to extract FLV metadata
- [x] Create a parser for the FLV metadata to extract the spatial reference of the FMV datasets that we could use for cataloging. **The two features of interest:**
  - [x] flight path
  - [x] footprint
- [x] Set up an ETL routine that will except FMV datasets and automatically do the cataloging
  - [x] Extracts FLV metadata with Kwiver and saves that in the Database
  - [x] parses features of interest from FLV data and saves those

banesullivan:
Resolve #142

Todo:

- [x] Use Kwiver Docker container to dump out FLV metadata
- [x] Attach FLV data to `FMVEntry` model
- [x] Implement ETL routines to populate flight path and footprints fields on `FMVEntry`
- [x] Should we choose a different name for the footprint polygons since `footprint` is already a field on `SpatialEntry`?
- [x] Add tests 

FYI: the Kwiver/TeleSculptor folks are working to improve the spatial metadata extraction from the videos so eventually, we should be able to leverage that and drop all of the regex parsings that I implement here.


utils/Dockerfile-kwiver-dump-klv
```diff
@@ -0,0 +1,14 @@
...
+#
+# Usage:
+#   cat video.mpg | docker run -i banesullivan/kwiver:dump-klv > video.klv
+FROM banesullivan/kwiver:latest
+
+# NOTE: vodeo file extension does not matter.
+CMD bash -c '\
+  cat > fmv.video && \
+  /kwiver/build/bin/kwiver dump-klv fmv.video \
+'
```
banesullivan, line 14: This is a Dokerfile for extracting KLV metadata from any FMV dataset. I wasn't sure where the best place to keep this Dockerfile would be, so I posted here and pushed the image to DockerHub [here](https://hub.docker.com/repository/docker/banesullivan/kwiver/tags?page=1) under the `dump-klv` tag. Perhaps this should migrate over to Kwiver down the road.

Also, note that we just pull the image and don't actually use this Dockerfile anywhere.

geodata/models/fmv/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __str__(self):
+        return f'{self.name} ({self.id})'
+
+    name = models.CharField(max_length=100, blank=True, null=True)
+    description = models.TextField(null=True, blank=True)
+
+    fmv_file = models.OneToOneField(FMVFile, null=True, on_delete=models.CASCADE)
+    klv_file = S3FileField(upload_to='files/fmv/', null=True, blank=True)
+    web_video_file = S3FileField(upload_to='files/fmv/web/', null=True, blank=True)
```
banesullivan, line 37: @manthey, how do you feel about this? We cannot display the raw videos in a web browser and these videos are significant in size, so I added a routine to use `ffmpeg` to convert the videos to a web-friendly format on import and save that video here (this is effectively duplicate data through all the spatial metadata in the video is lost during this conversion)

banesullivan:
The current FMV files we have are too large for testing. I am working with some others to produce a small testing set and the timeline is unknown. I can verify that this is working locally and I have added a test that skips the `FMVFile` upload process and pulls a premade KLV metadata file that I [have hosted on `data.kitware.com`](https://data.kitware.com/#item/5f61144e50a41e3d198932a5). This tests all the other parts of the ETL routine for parsing that metadata and populating an `FMVEntry` spatial entry, just not the extraction of that spatial metadata from the video

dev/django.Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apt-get update && \
         libmagic1 \
         libgl1-mesa-glx \
         libglib2.0-0 \
+        ffmpeg \
```
banesullivan, line 13: For converting the videos to a web-friendly format

geodata/models/fmv/base.py
manthey: I think it is the right thing to do.  I'm wondering if we should optimize the video for web playback as well as converting its format.  That is, for easy seeking, it is best to have a gop so that there is a keyframe every half second or so.  Similarly, we may want to have a maximum bitrate and other parameters.  And, using the copy codec is fast, but not guaranteed that it will play on a web browser. 

Other options might also be appropriate (e.g., `ffmpeg -i <source> -map_metadata -1 -vcodec libx264 -level 30 -maxrate 1024k -movflags faststart -g 15 -an <output>`)

manthey:
> The current FMV files we have are too large for testing. I am working with some others to produce a small testing set and the timeline is unknown. I can verify that this is working locally and I have added a test that skips the `FMVFile` upload process and pulls a premade KLV metadata file that I [have hosted on `data.kitware.com`](https://data.kitware.com/#item/5f61144e50a41e3d198932a5). This tests all the other parts of the ETL routine for parsing that metadata and populating an `FMVEntry` spatial entry, just not the extraction of that spatial metadata from the video

Can you use ffmpeg with -map_metadata to subset an existing file?

banesullivan:
> Can you use ffmpeg with -map_metadata to subset an existing file?

I have been trying/debugging doing that for quite some time today. I have not been able to preserve the metadata stream as I don't think it is actually timestamped in the stream - i.e. if I trim the video the metadata stream does not properly line up with the trimmed video (and I can't get the stream to copy over at all).

I talked with Matt L and Chet N about this and they are having the same issue on another project. They are working to create a tool to do this trimming and will have a few small test videos for us to use.

geodata/models/fmv/base.py
banesullivan: I updated the conversion to use those options and it drastically reduced the video file size

manthey: The one that most affects size is `-maxrate`.  The value I picked will work on even very slow connections, but I could see wanting higher quality at the cost of not working on slow connections.  DVDs, for example, have a typical average bitrate of 6144k (if I remember correctly, excluding audio) and a typical maximum rate of 10000k.  Here we specify the maximum rate; you can also specify more parameters to try to target an average rate via `-vb`.  So typical DVD quality would be `-vb 6144k -maxrate 10000k`.  And, for what it is worth, smaller gop (`-g`) will result in larger files that seek better.

banesullivan: @manthey, is this changing the rate and thus the number of frames in the output video? We may want to keep the same number of frames so that we can correspond those frames to the stored footprint polygons more easily for #146. Another approach would be to store the timestamp of the video with the footprint polygons to avoid worrying about the number of frames and to make it easier to implement #146 based on time rather than frame

banesullivan:
I've uploaded a testing video to data.kitware.com [here](https://data.kitware.com/#item/5f62539650a41e3d198e0021). The spatial metadata is incorrect but for testing purposes, that does not matter. I will now add a test that runs the full ETL routine with that sample video

banesullivan:
The test I implemented is not working... I am getting a strange error from Docker:

```
File "/opt/django/geodata/models/fmv/etl.py", line 25, in _extract_klv_with_docker
    client = docker.from_env(version='auto', timeout=3600)
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/client.py", line 84, in from_env
    return cls(
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/client.py", line 40, in __init__
    self.api = APIClient(*args, **kwargs)
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/api/client.py", line 188, in __init__
    self._version = self._retrieve_server_version()
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/api/client.py", line 212, in _retrieve_server_version
    raise DockerException(
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

Perhaps we have to do something special to run using Docker in the tox testing env?

geodata/models/fmv/base.py
manthey: No, this isn't changing the number of frames.  It is recompressing the video and changing which frames are key frames.

There are videos that have non-constant frame rates.  If we ever need to know the precise time of each frame, we'd have to store the frame time with each footprint.  In practice, most videos from cameras are constant frame rate (as opposed to videos made of power-point presentations).

manthey:
> Perhaps we have to do something special to run using Docker in the tox testing env?

We either have to use `setup_remote_docker` or a machine executor in CircleCI to create docker containers as part of tests.

banesullivan:
@manthey, assuming the CI checks come back green, I am going to merge this and #145 today and figure out the testing issue in a follow-up PR

PR #144, opened 2020-09-10
Make search viewer reusable

banesullivan:
This makes the viewer reusable. My goal in doing this is to give us one single viewer that we can place on any page and it will automatically display the search parameters and data footprints that are loaded from the page's context

This will enable us to more easily define any type of seach-by page for any model. I am doing this to have a search page for FMV over in #143 

Currently, this assumes the search parameters are a single point but when we add other search parameters, we can update this to handle those cases.

### DIFF
--- a/geodata/search.py
+++ b/geodata/search.py
@@ -123,14 +123,14 @@ def search_near_point_geometry(request, *args, **kwargs):
     return JsonResponse(serializers.GeometryEntrySerializer(results, many=True).data, safe=False)
 
 
-def extent_summary(found):
+def extent_summary_spatial(found):
     """
-    Given a query set of items, return a result dictionary with the summary.
+    Given a query set of SpatialEntry, return a result dictionary with the summary.
 
     :param found: a query set with SpatialEntry results.
     :returns: a dictionary with count, collect, convex_hull, extent,
-        acquisition, acqusition_date, created, modified.  collect and
-        convex_hull are geojson objects.
+        acquisition, acqusition_date.  collect and convex_hull are geojson
+        objects.
     """
     if found and found.count():
         summary = found.aggregate(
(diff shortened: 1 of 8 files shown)


geodata/search.py
```diff
@@ -123,27 +123,21 @@ def search_near_point_geometry(request, *args, **kwargs):
...
             Collect('footprint'),
             Extent('footprint'),
             Min('acquisition_date'),
             Max('acquisition_date'),
-            Min('created'),
-            Max('created'),
-            Min('modified'),
-            Max('modified'),
-            acquisition__min=Min(Coalesce('acquisition_date', 'created')),
-            acquisition__max=Max(Coalesce('acquisition_date', 'created')),
```
banesullivan, line 146: @manthey, I separated this extents summary function to handle `ModifiableEntry` and `SpatialEntry` separately. A consequence is loosing this coalesced field. We could do this manually in `extent_summary` but I don't see an immediate need for it

banesullivan:
Now there is a new `<host>/geodata/spatial_entries/` page that will list all spatial entries and show their `footprint`s on a map

geodata/search.py
manthey: In the endpoints, if we ask to query by acquisition time, and no `acquisition_date` is set, it falls back to using `created`.  The point of this coalesce was to report the min/max of this fallback field.  Can this be moved to the ` extent_summary_modifiable` function?

banesullivan: Ah I see, how about inside the new `extent_sumary` function which merges the results of `extent_summary_modifiable` and `extent_summary_spatial`? That's the only place where we can guarantee that both `aquistion_date` and `created` are present

manthey: Yes, that would work.

geodata/search.py
```diff
@@ -178,14 +192,29 @@ def extant_summary(found):
...
+    results.update(extent_summary_spatial(found))
+    if found and found.count():
+        summary = found.aggregate(
+            acquisition__min=Min(Coalesce('acquisition_date', 'created')),
+            acquisition__max=Max(Coalesce('acquisition_date', 'created')),
+        )
+        results['acquisition'] = [
+            summary['acquisition__min'].isoformat(),
+            summary['acquisition__max'].isoformat(),
+        ]
```
banesullivan, line 206: @manthey, I think this should do it

manthey:
@banesullivan I've made a commit that fixes zooming the to extents of the selected data.  If you like what I've done there, go ahead and merge this.

manthey (approved):

banesullivan:
Thanks, @manthey!

PR #145, opened 2020-09-10
improve front end searching

### LINKED ISSUE #146: FMV Detail View: show footprint of frame while previewing video
Task: show a live footprint on a map while playing the video.

#145 adds a detailed view page for FMV datasets that shows a preview of the video and the entire video's ground footprint. We need to update this page to synchronize the GoeJS viewer to show the current polygon for the frame shown in the video preview.

Unfortunately, the number of footprint polygons that we store is not guaranteed to have a 1-to-1 mapping back to the video frames (some frames could be lacking a footprint... in that case, we will use the last available footprint).

Because of that lack of 1-to-1 correspondence, we need a way to track which polygons correspond to which frames. At present, this information is only known during the ETL routine and the mapping is lost after we save the `FMVEntry`. To track this, we could store an array of the frame indexes for each of the Polygons that were saved to the model as a binary blob. Then we'd have an N-length array of those indices and a `MultiPolygon` of N polygons that we could pass to the view. One thing to note is that this could easily approach 10K polygons. From there we could have a custom video viewer coupled to the GeoJS viewer that will keep track of the current video frame and use that to update the polygon rendered by the GeoJS viewer.

@manthey, I'm not sure how to create this custom viewer that will track the frame index, any suggestions?

banesullivan:
Uses #144 and extends #143 

Resolve #146 resolves UI parts of #147 

This is a general overhaul of the frontend searching features and detail views for spatial entries.

Searching for spatial entries:

| BBox | Point+Radius |
|---|---|
| Screen Shot 2020-09-22 at 2 59 10 PM | Screen Shot 2020-09-22 at 2 57 26 PM |


geodata/templates/geodata/fmventry_detail.html
```diff
@@ -0,0 +1,28 @@
...
+      <h2 class="article-title">
+        <span>{{ object.name }} ({{object.pk}})</span>
+      </h2>
+      <hr/>
+
+      <p class="article-content">Description: {{ object.description }}</p>
+
+      <video width="320" height="240" controls>
+        <source src="{{ object.fmv_file.file }}" type="video/mpg">
+      </video>
```
banesullivan, line 20: Is it possible to show the video from the S3FileField?

geodata/templates/geodata/fmventry_detail.html
```diff
@@ -0,0 +1,28 @@
...
+      <h2 class="article-title">
+        <span>{{ object.name }} ({{object.pk}})</span>
+      </h2>
+      <hr/>
+
+      <p class="article-content">Description: {{ object.description }}</p>
+
+      <video width="320" height="240" autobuffer="autobuffer" autoplay="autoplay" loop="loop" controls="controls">
+        <source src="{{ object.fmv_file.file.url }}" type='video/mp4; codecs="avc1.42E01E, mp4a.40.2"'>
+      </video>
```
banesullivan, line 20: This is giving me the correct URL to the video source (I can inspect the page and download the video from that link). But it isn't playing (videos are in `.mpg` format)

banesullivan: So turns out these VIRAT videos are in `MPEG-TS` format which we may need special tooling to play in a browser. https://github.com/video-dev/hls.js/ may be able to handle transcoding the video

geodata/search.py
```diff
@@ -120,7 +120,7 @@ def search_bounding_box_filter(params):
     query = Q()
     if all(
         params.get(key) is not None
-        for key in {'minimum_longitude', 'minimumlatitude', 'maximum_longitude', 'maximum_latitude'}
+        for key in {'minimum_longitude', 'minimum_latitude', 'maximum_longitude', 'maximum_latitude'}
```
banesullivan, line 123: @manthey, FYI in case your testing branch conflicts

PR #154, opened 2020-09-28
Pin celery and djangorestframework.

manthey:
We need to make some changes to work with celery 5.  drf-yasg needs to make changes to work with drf 3.12.


brianhelba:
@manthey As this sort of thing happens in the future, it'd also be very helpful to make a corresponding PR at https://github.com/girder/cookiecutter-django-girder , to notify all projects with the same code ancestry of the issue.

PR #157, opened 2020-09-29
Move app modules under base project module

brianhelba:
This moves the `core` and `geodata` apps to be nested within the `rgd` base project. This significantly simplifies project packaging for deployment via pip, as all source code is within the `rgd/` directory.


brianhelba:
@manthey @banesullivan This is slightly disruptive to existing PRs, since almost every file is renamed. However, it's my experience that Git's merging logic is actually pretty smart about resolving this sort of conflict automatically.

Since `master` was previously broken due to upstream dependency issues, now is perhaps a good time to make this change, since all outstanding branches need to be merged with `master` anyway.

banesullivan (approved):
This sounds good to me! My only thought is on whether one day we want to split the geodata app out into its own repository so it can more easily be incorporated in other sites without the `core` app. Either way, these changes are good and we can always do that split later on if needed

banesullivan:
> This is slightly disruptive to existing PRs, since almost every file is renamed. However, it's my experience that Git's merging logic is actually pretty smart about resolving this sort of conflict automatically.

Yeah, this will be fun to merge into existing PRs but since you used `git mv` it should be clean and easy

PR #158, opened 2020-09-30
Improve the default name for files.

manthey:
Before, this ends up with a hash-prefix from minio.  This just uses the basename.


banesullivan (approved):
Ah nice! This was bugging me as well

PR #164, opened 2020-10-01
Load Bulma from a CDN, instead of vendoring it in the repo

brianhelba:
The recent file reorganization resulted in the Bulma CSS file being excluded. Rather than re-adding it to the repo, it's more maintainable to load it from a CDN.


banesullivan (approved):
Aha, this fixes the issue I saw. Thanks, @brianhelba!

PR #166, opened 2020-10-01
Backport gitignore from cookiecutter-django-girder

brianhelba:
This further unifies this app with the cookiecutter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,24 @@
-build/
+staticfiles/
 
-# Byte-compiled / optimized / DLL files
+# Created by https://www.gitignore.io/api/django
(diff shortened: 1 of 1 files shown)


manthey:
I guess it depends on the logic of adding all this stuff to `.gitignore`.  If we are trying to cover all the cases of random files that we don't want, it might be better to explicitly allow files rather than have this laundry list of files.  Since our developers often use vim, emacs, or vscode, if we are going to use gitignore.io, I guess I'd rather see those included, too (e.g., `https://www.toptal.com/developers/gitignore?templates=django,vim,emacs,visualstudiocode` as the base)

PR #168, opened 2020-10-01
Pin django-s3-file-field

brianhelba:
This is in anticipation of an upcoming `django-s3-file-field` release, which may introduce breakages.


brianhelba:
@manthey @banesullivan This shouldn't change anything. Just need an approval.

PR #172, opened 2020-10-01
Testing improvements

banesullivan:
- The KLV test was a bit out of control. I updated the example data on DKC to improve the `test_populate_fmv_entry_from_klv_file` test from 32.88s, 69.2 MB to 4.71s, 14.6 MB.
- This also adds memprof to the pytest env to output memory consumption to keep us on our feet moving forward
- output test durations as well
- Decreased warnings from 60 to 4
    - Adjust the search query to avoid the recurring warning: `RuntimeWarning: DateTimeField ... received a naive datetime (2020-10-02 21:10:34) while time zone support is active.`
    - Ignore `RemovedInDjango40Warning` as we are pretty set on sticking with Django 3.x for now and these seem to all be coming from s3filefield/auth 
    - Note: left over warnings are coming from external libraries like `django-minio-storage` and `configurations`. Also one from rasterio because we are using it to read general imagery without a spatial reference
- Decreases pytest execution from ~62s to ~32s (at least locally)


rgd/geodata/search.py
```diff
@@ -66,6 +67,8 @@ class BoundingBoxSerializer(rfserializers.Serializer):
 
 
 def _add_time_to_query(query, timefield, starttime, endtime, has_created=False):
+    starttime = make_aware(starttime)
+    endtime = make_aware(endtime)
```
banesullivan, line 71: This could have unintended side effects... perhaps we need to extend the endpoint to accept a DateTime with timezone? or if it does have a timezone, skip this?

tox.ini
```diff
@@ -55,8 +55,9 @@ deps =
     pytest-factoryboy
     pytest-mock
     pytest-cov
+    pytest-memprof
 commands =
-    pytest --cov=rgd --cov=core --cov=geodata {posargs}
+    pytest --cov=rgd --cov=core --cov=geodata --durations=0 {posargs}
```
brianhelba, line 60: Consider adding additional CLI options to the `[pytest] addopts` section if you want them to always be included.

banesullivan: Done

banesullivan:
Merging... if issues arise, we can adjust later. These testing improvements will help me move faster on some other work

PR #177, opened 2020-10-09
[WIP] Dev environment notes

mcovalt:
This PR addresses a newcomer's issues setting up the dev environment.

### Environment vars

- [[a892b4a]](https://github.com/ResonantGeoData/ResonantGeoData/commit/a892b4a09995a6664020386cbc225fbf618711ee) The `DJANGO_MINIO_STORAGE_MEDIA_URL` was not set for the native environment variables. [~Also, it seems like the equivalent environment variable is not set correctly for the docker-compose network hostname?~](https://github.com/ResonantGeoData/ResonantGeoData/pull/177#discussion_r503321255)
- Is there a wiki/document with a description of each environment variable? They're all pretty self explanatory, but it did take a second to look into each one.
- Any thoughts on standardizing to a ".env file" at the root of the repo with actual environment variables taking precedence? It's nice since many tools also do this ([IDEs](https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file), [docker-compose](https://docs.docker.com/compose/environment-variables/#the-env-file), etc.).

### Developing natively

The approach to developing natively is pretty open.

- There's the approach to have [tox setup your development environment](https://tox.readthedocs.io/en/latest/example/devenv.html). That seems pretty cool and goes along with the approach to just run `pip install -e .[dev]` as stated in the "README.md" since [`[dev]` only contains tox and ipython](https://github.com/ResonantGeoData/ResonantGeoData/blob/master/setup.py#L45). This would need to be implemented in the "tox.ini", though.
- Otherwise the virtual environment will need to be installed with `pip install -r requirements.txt -e .[dev]`. But where?
  - [Black](https://github.com/ResonantGeoData/ResonantGeoData/blob/master/pyproject.toml#L5) and [.gitignore](https://github.com/ResonantGeoData/ResonantGeoData/blob/master/.gitignore#L50) are ignoring ".venv". But Flake8 is trying to lint it.
  - [Flake8](https://github.com/ResonantGeoData/ResonantGeoData/blob/master/tox.ini#L69) is ignoring "env38". But .gitignore isn't.

### Developing in a container

This is pretty cool. I always enjoy the idea of perfect dev/prod parity (especially with so many compiled binary Python modules), but usually hit some roadblock so I typically avoid. Any thoughts on [VS Code's native support for it](https://code.visualstudio.com/docs/remote/containers)? I like it, but the setup is a bit intense. We could add `.vscode/devcontainer.json` if someone ever gets it setup.

### Dependencies

- **GDAL**: I'd like to see the build process. Where can I find the script that was used to build the wheels?
- **ffmpeg**:  Are we going to need to all be on the same version?


manthey:
@mcovalt The GDAL wheels (and a variety of other wheels) are built with a very ugly Dockerfile here: https://github.com/girder/large_image_wheels

dev/.env.docker-compose
```diff
@@ -5,4 +5,4 @@ DJANGO_MINIO_STORAGE_ENDPOINT=minio:9000
 DJANGO_MINIO_STORAGE_ACCESS_KEY=minioAccessKey
 DJANGO_MINIO_STORAGE_SECRET_KEY=minioSecretKey
 DJANGO_STORAGE_BUCKET_NAME=django-storage
-DJANGO_MINIO_STORAGE_MEDIA_URL=http://localhost:9000/django-storage
+DJANGO_MINIO_STORAGE_MEDIA_URL=http://minio:9000/django-storage
```
manthey, line 8: I'm not sure this is actually desired.  Specifically, `minio` is the host name within the docker network, but won't work outside of the docker network (e.g., from a browser on localhost), whereas if the port is forwarded, it will work on localhost (but not from an external machine).

manthey: In my deployment, I've got ` DJANGO_MINIO_STORAGE_ENDPOINT=<hostname of my machine>:<port>`, where I've used the actual host name on our VPN.  This lets me access the application from another machine in or vpn and still upload and download to minio.

mcovalt: I see. I did not realize this is for manufacturing the URL for clients.

manthey: I could be wrong, as I might be confusing settings we had earlier on in the project.  Did you check if uploads and downloads both work from the local machine with `minio` in the url?

mcovalt: I just checked and can confirm that the change I proposed here would break the uploads/downloads. For my future reference, the `DJANGO_MINIO_STORAGE_MEDIA_URL` environment variable [is used to set the "settings.py" variable `MINIO_STORAGE_MEDIA_URL`](https://github.com/girder/django-composed-configuration/blob/8c210a44046c691f28c358b4f3efe868e0a2b03b/composed_configuration/_configuration.py#L54-L56) for the package [django-minio-storage](https://github.com/py-pa/django-minio-storage). The setting is documented [here](https://django-minio-storage.readthedocs.io/en/stable/usage/). This variable is used to create the URLs provided to clients for files that live in minio. It's analogous to the [stock Django `MEDIA_URL` setting](https://docs.djangoproject.com/en/3.1/ref/settings/#media-url).

banesullivan:
@jtomeck, some of the notes here may be helpful when getting the site up and running locally

PR #179, opened 2020-10-09
Fix issues with failure_reason and status when executing tasks

### LINKED ISSUE #167: improve status flag for TaskEventMixin
Currently, there is a status property visible in the admin interface that basically just checks if the `failure_reason` has content or not. This is way too basic to provide much insight into the *status* of the model and it is difficult to know the status of a model via the admin interface without looking at the logs.

To do: improve the `TaskEventMixin` model to have an internal way of monitoring the associated task and displaying whether that task is one of these states:

https://github.com/ResonantGeoData/ResonantGeoData/blob/cda3fcf38a7d9134bd62ddabbfaa3f9fc3d5e602/rgd/core/models.py#L217-L223

This is exactly in line with what was done in the `core` app for `AlgorithmJob`. 

I will update the `TaskEventMixin` like this but I'm not sure if it makes sense to share this model with the core app or not (considering the core app isn't really being developed or used lately)

### LINKED ISSUE #169: improve failure_reason for TaskEventMixin
Currently, most models that mixin the `TaskEventMixin` only add the final error message to their `failure_reason`. We should improve those routines to capture a full traceback when possible.

Also, it would be good to capture the log much like what is done for `AlgorithmResult` in the `core` app

banesullivan:
Resolves #169: Reworked the execution of the tasks to have a standard way of capturing a full traceback of the failure reason. Note that the `failure_reason` is only set by the task executor and not the ETL routines themselves. The ETL routines through an error directly if called.

Resolves #167: This adds a new `status` field for all models in `geodata` that use `TaskEventMixin`

requires #186


banesullivan:
I am going to go ahead and merge this when the checks are green as I need to do another migration and it's just way easier to do that linearly rather than across branches

rgd/geodata/models/mixins.py
```diff
@@ -17,6 +27,12 @@ def _run_task(self):
         """Validate the raster asynchronously."""
         if not callable(self.task_func):
             raise RuntimeError('Task function must be set to a callable.')  # pragma: no cover
+        self.status = Status.QUEUED
+        # self.save(
+        #     update_fields=[
+        #         'status',
+        #     ]
+        # )
```
banesullivan, line 35: See https://github.com/girder/django-s3-file-field/issues/105#issuecomment-707922293

banesullivan:
Note that this is good to go off of #186

mcovalt (approved):
LGTM!

PR #182, opened 2020-10-11
GeoJSON based search

### LINKED ISSUE #152: Add search endpoints where an arbitrary geojson object can be used for the search area
We may also want an option on whether the data needs to intersect the object or be strictly within it.

This would be similar to the bounding box search for time handling/searching.

mcovalt:
Closes #152 

This adds the following endpoints allowing search to be performed via a GeoJSON geometry:

- /api/geodata/geojson
- /api/geodata/raster/geojson
- /api/geodata/geometry/geojson
- /api/geodata/geojson/extent
- /api/geodata/raster/geojson/extent
- /api/geodata/geometry/geojson/extent

The endpoints take a query parameter `geojson` which should be set to the URL encoded JSON string of a GeoJSON feature.

Capture

Note: the `geojson` query parameter doesn't necessarily have to be a GeoJSON feature. It could also be a WKT/EWKT or a HEX/HEXEWKB [due to the flexibility of the `GEOSGeometry` class](https://docs.djangoproject.com/en/3.1/ref/contrib/gis/geos/#geosgeometry). This will be well documented for the unified search/facet endpoint which will take a more generic `q` parameter.


rgd/geodata/templates/geodata/_includes/search_form.html
```diff
@@ -73,6 +74,16 @@
     </form>
   </div>
 
+  <div id="GeoJSON" class="tabcontent">
+    <form method="GET">
+      GeoJSON geometry: <input type="text" name="geojson" value="{{ request.GET.geojson }}">
+      <br>
+      within: <input type="checkbox" {{ request.GET.within|yesno:"checked,," }} name="within">
```
banesullivan, line 81: Let's add a little tooltip

```suggestion
      within: <input type="checkbox" {{ request.GET.within|yesno:"checked,," }} name="within" title="Only retrieve entries that are fully within the boundary of the feature.">
```

rgd/geodata/templates/geodata/_includes/search_form.html
```diff
@@ -58,7 +59,7 @@
     </form>
   </div>
 
-  <div id="Boundung Box" class="tabcontent">
+  <div id="Bounding Box" class="tabcontent">
```
banesullivan, line 62: Good catch! Thanks!

banesullivan (approved):
This looks great! I have one, very minor suggested addition

My only concern right now is what happens when the query is invalid. At the moment, a server error occurs. It'd be good to utilize the alert message template we have in the `base.html` to re-render the page with an error message, but that might require a separate PR with internal changes to the views.

Further, I'd like to add a way to draw a polygon on the GeoJS map and have that populate the form. I will open a new issue for this (assigned to myself).

mcovalt:
> It'd be good to utilize the alert message template we have in the base.html to re-render the page with an error message.

I agree. I'll look into implementing that.

PR #183, opened 2020-10-12
Remove some unnecessary usage of _field_file_to_local_path

brianhelba:
S3FileField behaves exactly the same as FileField once files have been uploaded.

For more information on how to use FieldFile and File objects, see: https://docs.djangoproject.com/en/3.1/topics/files/#the-file-object for more

### DIFF
--- a/rgd/core/tasks.py
+++ b/rgd/core/tasks.py
@@ -35,19 +35,19 @@ def _run_algorithm(algorithm_job):
             except docker.errors.ImageNotFound:
                 pass
         if not image:
-            with _field_file_to_local_path(algorithm_file) as algorithm_path:
-                logger.info('Loading docker image %s' % algorithm_path)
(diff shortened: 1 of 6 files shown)


brianhelba:
There are a few more places where `_field_file_to_local_path` could still likely be removed, but were too complicated for me to confidently refactor.

For some libraries like `kwcoco` and `gdal`, a Python file-like object is not accepted as input. It is *only* those places where `_field_file_to_local_path` is actually necessary.

Otherwise, it is simpler and more efficient to interact with a `FileField` (i.e. an instance of a `FileField` field) using its native `.open` method to return a file-like object.

banesullivan (approved):
> There are a few more places where _field_file_to_local_path could still likely be removed, but were too complicated for me to confidently refactor.

I can handle those in a follow up

PR #188, opened 2020-10-16
Styles cleanups

jtomeck:
This pull request will implement SASS as well as some general style cleanups sitewide.


docker-compose.yml
```diff
@@ -28,6 +28,11 @@ services:
     volumes:
       - fsdata:/data
 
+  sass:
+    image: larryprice/sass
+    volumes:
+      - rgd/core/static/css/sass:rgd/core/static/css/
```
banesullivan, line 34: I think this needs to mount to `:/src` for the `larryprice/sass` image to compile it

I'm testing locally.........

docker-compose.yml
```diff
@@ -28,6 +28,11 @@ services:
     volumes:
       - fsdata:/data
 
+  sass:
```
brianhelba, line 31: If this is going to merged, I'd suggest making these changes to `docker-compose.override.yml` instead. Developers running the Python code locally may want more control to run the SASS watcher locally too.

banesullivan: Sounds good to me, thanks Brian!

banesullivan:
@jtomeck, I'm curious if you've made any progress on the final few items we discussed? For repitition, I'm putting the checklist here:


- [ ] Can we improve the layout of the detail pages?
- [ ] Switch tabs to have Geospatial and Faceted search forms
	- [ ] Move current geospatial tabs to hide/show with dropdown choice
- [ ] Put stats (count) in the top bar
- [ ] Make the table better/scrollable (not running off page)
- [ ] Copy/paste button for the GeoJSON field in detail view only
- [ ] Remove Welcome button and get rid of drop down

jtomeck:
@banesullivan - a couple things about this. I couldn't get the button copying to work properly. I also think it should display an alert that the text copied successfully.

Also, I think there needs to be an if statement in the titlebar to only show the meta information if you're on the detail page. Everything I tried wasn't working so I thought I'd stop wasting time trying to figure out things that're probably trivial for you.

banesullivan:
> Also, I think there needs to be an if statement in the titlebar to only show the meta information if you're on the detail page. Everything I tried wasn't working so I thought I'd stop wasting time trying to figure out things that're probably trivial for you.

I went ahead and handled this by implementing it a bit different. The base template imports the block and the sub templates for search and detail views add specific content per what makes sense.

It looks good in the detail view. But the search view has the "Spatial Entries" column in a weird place...

| Detail View | Search View |
|---|---|
| Screen Shot 2020-11-11 at 10 37 31 PM | Screen Shot 2020-11-11 at 10 37 37 PM |

---- 

> I couldn't get the button copying to work properly.

Hm. I'll see if I can do some research on how to implement this if you don't have time to hook it up. Having the styling and the button enough for me to take it from there

>  I also think it should display an alert that the text copied successfully.

ah yes, absolutely! Could you hook up some sort of alert/dialog/response? Once we figure out how to actually copy that content to the clipboard, we can do a status check of some sort before displaying it. For now, maybe just have an alert that appears any time the button is pressed?

jtomeck:
Can you explain in a little further detail what's happening with the tabs? I understand that there will be 2 tabs (Geospatial and Faceted Search) and that there will be a dropdown with the current tabs living in it. My questions are:

1. Is there any content or form already built that you want me to include on each of the 2 tabs?
1. Where does the dropdown live now? Is it under only 1 of the tabs, or is it a general thing that will need to be displayed regardless of what tab you're on?
    - My reason for asking this is because we might want to keep it in it's own panel if the information in the dropdown needs to be global

banesullivan:
Going to answer backward because I think that makes sense to explain what we are looking for.

> 2. Where does the dropdown live now? Is it under only 1 of the tabs, or is it a general thing that will need to be displayed regardless of what tab you're on?

This dropdown is not global but within the geospatial search tab. The dropdown is for choosing which type of geospatial search to perform, with options: `Point Radius`, `Bounding Box`, and `GeoJSON` for the three search forms we currently have in `rgd/geodata/templates/geodata/_includes/search_form.html`

> 1. Is there any content or form already built that you want me to include on each of the 2 tabs?

Yes for the geospatial tab and no for the faceted search tab. 

The dropdown in the geospatial tab will hide/show the fields from the three current forms we have based on the type of search (as mentioned above). For the faceted search tab, can you add a form with some arbitrary fields for us?

When implementing the form in the Geospatial tab, can you have *one* form, where the dropdown is contained within to hide/show specific fields, then have the `Submit` button outside that dropdown context but still in the form? For now, the `Use Map` button with varying callbacks can be hidden/shown with its corresponding fields.

Does this provide enough context?

banesullivan:
@jared, on the latest changes, I noticed that the "Copy GeoJSON" button overlaps the metadata columns in the title bar

Screen Shot 2020-11-12 at 8 07 57 AM

jtomeck:
> @jtomeck, on the latest changes, I noticed that the "Copy GeoJSON" button overlaps the metadata columns in the title bar
Oof, smaller screen size. I also don't have anything showing up under created or modified, so it was hard to plan the widths of that. I'll figure it out!

jtomeck:
@banesullivan let me know how this looks now.

banesullivan:
There is still something going on with the vertical alignment. See screen capture. Can we fix this so the page has no vertical scroll?

banesullivan:
Also, there is still some minor overlap with the Copy GeoJSON button. Maybe we should move it to overlay on the map? (I can handle that positioning)

Screen Shot 2020-11-15 at 11 09 16 AM

banesullivan:
Otherwise, this is looking excellent and I can handle the rest of the tweaking/getting things working.

rgd/geodata/static/geodata/css/sass/_base.scss
```diff
@@ -11,6 +11,7 @@ html {
   font-weight: 300;
   height: 100%;
   margin: 0;
+  overflow: hidden;
```
jtomeck, line 14: @banesullivan - I'm not sure what the little scroll issue is. I've gone through every element in dev tools and made sure the padding and margins were set properly and that there was no additional element or pseudo element causing the page to be slightly higher.

This is a bit of a hacky solution, but it works. Let me know if you have any issues with it.

banesullivan: Huh. very odd. For now, this works for me

PR #190, opened 2020-10-20
Move signals to dedicated file

mcovalt:
This PR moves all signals to a dedicated "signals.py" file for each app. Then, using each app's [`AppConfig.ready()`](https://docs.djangoproject.com/en/3.1/ref/applications/#django.apps.AppConfig.ready), imports that file. This guarantees registration of all signals.

The `AppConfig` for each app was not being registered. The `INSTALLED_APPS` setting was updated to allow Django to find it. Read about this [here](https://docs.djangoproject.com/en/3.1/ref/applications/#configuring-applications).

### DIFF
--- a/rgd/core/__init__.py
+++ b/rgd/core/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'rgd.core.apps.CoreConfig'
--- a/rgd/core/__init__.py
+++ b/rgd/core/__init__.py
@@ -1,1 +1,0 @@
-default_app_config = 'rgd.core.apps.CoreConfig'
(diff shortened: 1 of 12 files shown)


manthey:
What were the side effects of not registering `AppConfig`?

mcovalt:
Prior to this PR, there were no side effects of not registering `AppConfig`. The `AppConfig` subclassing allows some customization of the app (e.g. the label in the admin panel). We were not using any of this customization before.

This PR uses the `AppConfig` subclassing to more reliably register signals. I had a case where a signal was not being registered because I did not import the relevant "models.py" file where the signal had existed. By hooking into `AppConfig.ready()`, we can initialize these signals at a safe part of the [application registration process](https://docs.djangoproject.com/en/3.1/ref/applications/#how-applications-are-loaded) every time the Django environment is created.

manthey (approved):

PR #192, opened 2020-10-28
Bump kwimage to remove torch dependency

banesullivan:
Jon Crall recently released a new version of `kwimage` and `kwarray` that makes `torch` and `pandas` optional dependencies. These changes should leverage that and significantly decrease the size of the virtual environment deployed on Heroku

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,5 @@
 # This is used by heroku for deployment
 --find-links https://girder.github.io/large_image_wheels
 GDAL
-# Heroku only uses CPU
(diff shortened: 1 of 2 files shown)


banesullivan:
I suppose to test this, we need to ssh into circle ci and make sure both torch and pandas are not installed. 

I'm not too sure how to compare the size difference of the virtual env... perhaps look at the size of master on Heroku then look at the size again after merging this?

banesullivan:
I ssh'd into CircleCI and confirm that torch is not installed but pandas still is. Pandas is likely being pulled from another one of our dependencies and this probably isn't too big a deal.

PR #193, opened 2020-10-28
Move video inside map UI layer for FMV detail page

banesullivan:
This is a minor change to move the FMV video onto the GeoJS map's UI layer.

Screen Shot 2020-10-28 at 10 25 35 AM

We cannot display the video over the red marker because it is a polygon. AFAIK, only quads support overlain video -- we cannot use a quad here because the video frame's projection on the ground is typically a trapezoid

### DIFF
--- a/rgd/geodata/templates/geodata/fmventry_detail.html
+++ b/rgd/geodata/templates/geodata/fmventry_detail.html
@@ -34,6 +34,10 @@
       let frameRate = JSON.parse('{{ frame_rate|escapejs }}');
       var video = document.getElementById("fmv_video");
 
+      // Move video to UI layer in map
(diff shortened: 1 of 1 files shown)


manthey:
Quads are arbitrary quadrilaterals.  That is, they include trapezoids.

banesullivan:
Ah. I totally did not realize that was the case in GeoJS. I thought they had to be rectangles aligned with the cartesian axes. To define a trapezoid with a quad, would I just specify all for coordinates instead of two opposing?

manthey:
Yes, you can specify all corners.  See https://opengeoscience.github.io/geojs/examples/quads/.  The corner designations (`ll`, `ul`, `ur,` `ul` for lower/upper-left/right) are for the image or video (that is lower-left can be above and to the right on a map -- it is the position of the lower-left corner of the image/video).

PR #195, opened 2020-10-28
Add annotation layer for autofilling GeoJSON search field

### LINKED ISSUE #194: Add polygon annotation layer to GeoJS map for new GeoJSON endpoint
Following #182, add a new annotation layer to the GeoJS map that can handle drawing single polygons and return GeoJSON of that polygon for the new search endpoint.

As a part of this effort, it may be cool to enable a user to drag 'n' drop a shapefile (or textfile containing GeoJSON) to populate the form

banesullivan:
Resolve #194

requires #182


banesullivan:
Merging. This currently supports only searching with one polygon at a time. We can improve in later PRs if desired

PR #197, opened 2020-10-30
Add non-null constraint to SpatialEntry and much refactor

### LINKED ISSUE #162: extent summaries fail on models that aren't updated yet
I'm finding that the extent summary methods will throw an error if any one of the models in the query are missing a field. If I kick off a task that starts populating a spatial entry but the footprint (for example) hasn't been filled in yet then when we try to get the `geojson` attribute of that field (which is `None`) an error is thrown and the search results are not returned and the frontend site shows an error page.

It seems we need a way to filter the query for models that have the fields being summarized actually populated. Otherwise, this can throw an error frequently if for example a user uploads a lot of FMV videos or raster datasets as those could take several minutes to populate.

### LINKED ISSUE #199: ImageEntry's image_file relationship needs a non-null constraint
Due to the way I implemented KWCOCO support, I let that relationship be nullable. This means an `ImageEntry` can become an orphan with no original `ImageFile` (in the case that the `ImageArchive` was deleted).

We need a way to batch upload images rather than uploading a zip/tar of many images so that each image gets an `ImageFileEntry`

banesullivan:
Resolve #162, Resolve #199 

This adds a new non-null constraint to the `SpatialEntry` model - a much-needed change! In order to do this, I had to significantly refactor the `RaterEntry` model to move the spatial info into a separate model that is automatically generated.

The migrations had to be redone for this so please reset your database after this merge

Other changes:

- move around some of the fields on the FMV models
- make fields that shouldn't be null also have a non-null constraint
- clean up usage of `blank=True` in fields (many fields had this and it was unnecessary)
- Change `ForeignKey` to `OneToOneField` where appropriate
- major improvements to the admin interface to clean up UI, navigation, and running ETL routines
    - New actions for creating `ImageSet` and `RasterEntry` from `ImageEntry` list
    - Use the inline admin interface to show autogenerated models on same page as the parent model (this declutters the main admin interface)
    - improve the readonly field on some models


rgd/geodata/search.py
```diff
@@ -14,7 +14,7 @@
 from rest_framework.decorators import api_view
 
 from . import serializers
-from .models import GeometryEntry, RasterEntry, SpatialEntry
+from .models import GeometryEntry, RasterMetaEntry, SpatialEntry
```
banesullivan, line 17: @manthey, @mcovalt: how do you all feel about this. The changes in this PR remove the spatial reference from `RasterEntry` and into `RasterMetaEntry` so that is what has to be queried against

mcovalt: This seems fine to me. Thanks for the heads up!

mcovalt:
Helps normalize the data. This will help a lot as we start building more complex queries!

mcovalt (approved):

banesullivan:
Thanks, @mcovalt! I squashed the migrations into one file and tested all of the ETL routines manually through the admin interface ...merging after one more minor tweak

PR #204, opened 2020-11-05
Follow up #197: fix front-end issues

banesullivan:
Follow up to #197. This fixes some front-end and minor ETL issues I left behind


rgd/geodata/models/imagery/base.py
```diff
@@ -104,6 +104,24 @@ def __str__(self):
...
+        return self.rastermetaentry.outline
+
+    @property
+    def count(self):
+        """Get number of bands across all images in image set."""
+        n = 0
+        for im in self.image_set.images.all():
+            n += im.number_of_bands
+        return n
+
```
banesullivan, line 124: For backward compatibility and making the views simpler. We can resolve later if this is bad practice

banesullivan:
I'm going to merge to move along with #188

PR #205, opened 2020-11-10
fix thumbnails and outline geometry

### LINKED ISSUE #176: Raster thumbnails not aligned correctly on GeoJS map
With the introduction of thumbnails in #175, there is a significant error in how we align the thumbnail on the map view. See screenshots:

Screen Shot 2020-10-08 at 1 38 55 PM

Screen Shot 2020-10-08 at 1 41 59 PM

I have confidence in the valid data footprint (red outline).

I suspect we just need to adjust the quad that holds the thumbnail. Note how in the top image, you can see a few roads in the image and on the underlying map and there is a seemingly linear offset.

banesullivan:
Resolve #176 

For future reference, use the following example rasters to verify that the thumbnails are working.

- `LC08_L1TP_033032_20200422_20200508_01_T1_pixel_qa.tif`
- `landcover_sample_2000.tif`
- `vegdri_emodis_week34_082320.tif`

Summary of change: 
- new admin actions to reprocess image entries and raster entries (calls `save` to trigger save event and recreate the thumbnail or spatial reference)
- added a helper method to reproject an open rasterio raster to a given ESPG
- During image thumbnail generation routine: reproject raster to Web Mercator before extracting if the image file has a spatial reference
- During raster outline/valid footprint calculation: 
  - first, reproject the raster to the DB's spatial reference (something was off when transforming the extracted geometry)
  - use rasterio/GDAL to extract a mask and create the GeoJSON feature (as opposed to our own convex hull approach)
- Raster detail view page now limits the view bounds on the map
- Update centroids in the raster tests since we are extracting the geometry properly now


banesullivan:
The thumbnail is still a bit off... I have confidence in the outline and the convex hull though:

| example | zoomed |
|---|---|
|  Screen Shot 2020-11-09 at 9 50 17 PM |  Screen Shot 2020-11-09 at 9 50 26 PM |

Note how the image shows the roads that should be next to the boundary, just not in the right spot. This is why I have confidence in the footprint/boundary and not the thumbnail

banesullivan:
Currently struggling to improve this any further. It works really well for most of the data in #221 but still has a slight issue with the larger rasters like for Landsat:

Screen Shot 2020-12-30 at 3 57 52 PM


or this continental-scale one:

Screen Shot 2020-12-30 at 4 01 06 PM

banesullivan:
For the really big raster, our outline is the same as what is shown on EarthExplorer, but I'm not sure how to correctly display the thumbnail:

Screen Shot 2020-12-30 at 4 19 03 PM

manthey:
Based on your continental image example, it looks like you aren't reprojecting the image when you make the thumbnail.  I think you have to either reproject the image into web mercator (to match the map) OR draw the map in the projection of the image.  For single thumbnails, changing the map projection might be easiest (e.g., https://opengeoscience.github.io/geojs/examples/reprojection/).

banesullivan:
> you have to either reproject the image into web mercator (to match the map)

I can reproject the raster to Web Mercator when extracting the thumbnail and get something a little better, but the geometry of the quad which the image is overlain is still incorrect.

Screen Shot 2021-01-04 at 9 09 38 AM

I tried transforming the geometry to ESPG:3857, but GeoJS throws an error with that geometry. i.e.:

```py
self.object.footprint.transform(3857, clone=True)
```

for `_SpatialDetailView` in `rgd/geodata/views.py`


> OR draw the map in the projection of the image. 

I also tried setting the projection of the GeoJS map to what we use in the database but it doesn't seem to work:

```js
let map = geo.map({
      node: '#map',
      clampBoundsX: true
    })
    basemapLayer = map.createLayer('osm', {
      source: 'osm',
      gcs: 'EPSG:4326'
    });
```

Screen Shot 2021-01-04 at 9 08 15 AM

banesullivan:
Aha! managed to get it right with some nuanced transforms. Something was going awry with our `transform_geometry` helper.

Screen Shot 2021-01-04 at 11 54 49 AM

Now I just need to confirm this is working for all our examples

banesullivan:
Local changes confirmed working with the original problematic raster:

Screen Shot 2021-01-04 at 11 57 47 AM

banesullivan:
@manthey, do you think I should cache these thumbnails on DKC and add a regression test to verify these stay consistent moving forward?

rgd/geodata/models/imagery/etl.py
```diff
@@ -35,6 +45,9 @@
 
 logger = get_task_logger(__name__)
 
+GDAL_DATA = os.path.join(os.path.dirname(rasterio.__file__), 'gdal_data')
+os.environ['GDAL_DATA'] = GDAL_DATA
```
banesullivan, line 49: Might need to also set a PROJ env variable for NTIF format issue in _reproject_raster

rgd/geodata/models/imagery/etl.py
```diff
@@ -67,6 +80,46 @@ def get_band(n):
     return ContentFile(byte_im)
 
 
+def _reproject_raster(src, epsg):
+    """Reproject an open raster to given spatial reference.
+
+    This will return an open rasterio handle.
+
+    """
```
banesullivan, line 88: add a check to see if open raster is already in that SRID and return it if so

rgd/geodata/models/imagery/etl.py
banesullivan: Will look into this later.

PR #208, opened 2020-11-23
Add faceted search

mcovalt:
This simply adds the [django-filter](https://django-filter.readthedocs.io/en/stable/) based faceted search API endpoint at `api/geodata/search`.

- `q`: geometry (sorts sets by NN)
- `predicate`: DE-9IM spatial predicate
- `relates`: DE-9IM string
- `distance` (min/max)
- `acquired` (min/max)
- `created` (min/max)
- `modified` (min/max)
- `datatype`: choice of raster, geometry, or FMV
- `instrumentation` (search with `icontains`)
- `num_bands` (min/max)
- `resolution` (min/max)
- `frame_rate` (min/max)

I can see a way where this could replace the [`search.py`](https://github.com/ResonantGeoData/ResonantGeoData/blob/master/rgd/geodata/search.py) module entirely. I had been working on implementing the GUI for this, but the code duplication was getting a bit out of hand. Instead, I am working on a single `SpatialEntryViewSet` using this filter. The drop-downs for GeoJSON, point, and bbox can all be implemented as widgets to populate `q`. However, this GUI based work is taking a back-burner on to-do in order to get permissions down before Dec 1.

### DIFF
--- a/rgd/geodata/api/__init__.py
+++ b/rgd/geodata/api/__init__.py
@@ -1,0 +1,9 @@
+"""Views for the API.
+
+Views routed under `/api/geodata`. The primary `Content-Type` recieved and served
+by these views ought to be `application/json`.
+"""
+
+__all__ = ['download', 'search']
+
+from . import download, search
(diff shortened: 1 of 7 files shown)


banesullivan (approved):
I'm ready to merge this

Issue #201, opened 2020-10-30
Add endpoint to serve ROIs of images as raw arrays


banesullivan:
(for the ndsampler use case)

We need to add endpoints to access arbitrary ROIs of images and serve the pixels of those ROIs as raw arrays.

This will likely require #198 under the hood

Will we want to use `large_image` for this?

-----

I imagine the endpoints being something like `GET:api/geodata/imagery/image_entry/roi/<ROI query parameters>` which will send back a serialized array of that ROI. The ROI query parameters will be a tricky part to get right for versatility. Immediately, I can think of a few common ways folks may want to use this:

- simple pixel coordinate bounding box: `umin,umax,vmin,vmax`
- geospatial bounding box: `xmin,xmax,ymin,ymax` (world coordinates)
- GeoJSON polygon we serve the region of the image surrounding that polygon

Perhaps each of those should be seperate endpoints?

- `GET:api/geodata/imagery/image_entry/roi?geojson=<geojson str>`
- `GET:api/geodata/imagery/image_entry/roi/pixel_box?umin=50&umax=150&vmin=90&vmax=200`
- `GET:api/geodata/imagery/image_entry/roi/geo_box?minimum_longitude=-107.35&maximum_longitude=-104.63&minimum_latitude=39.48&maximum_latitude=41.51`

banesullivan:
@manthey, thoughts?

Also, how should I approach serving the raw data? Should we use this as a motivator to integrate `large_image`?

manthey:
@banesullivan Do we know what format we want to server the raw arrays in?  large_image serves image tiles; currently these can be in PNG, JPEG, or TIFF.  The PNG and TIFF are lossless, but I think they are always reduced to 8 bits-per-sample.  One virtue of large_image is that is already handles reprojections.  If needed it would be straightforward to add an additional output format (protobuf or something similar, for instance, or TIFFs with a different bits-per-sample).

Erotemic:
> each of those should be separate endpoints

Yes

>  and serve the pixels of those ROIs as raw arrays.

Ultimately, I'll want to get a numpy array of pixel values, but the if we are talking about serialization for network transport, I don't think raw arrays is the right choice. I think its perfectly reasonable for RGD to send some encoded subimage region and then assume that the user is able to do something like calling `cv2.imdecode` or some GDAL equivalent on the result.

PR #209, opened 2020-12-01
Add COG conversion endpoint

### LINKED ISSUE #198: COG ImageEntry conversion endpoint
We need to add an endpoint to convert any ImageEntry to COG format

banesullivan:
Resolve #198 

Adds two new endpoints: 
- `api/geodata/imagery/image_entry/convert/<int:pk>`: create a new converted image in COG format
- `api/geodata/imagery/converted_image_files/<int:pk>`: Check status of converted image entry

To do

- [x] fix `gdal_translate` in PATH on CircleCI
- ~change test to validate that a COG is generated~
    - https://pypi.org/project/tifftools/
    - https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/samples/validate_cloud_optimized_geotiff.py
- [x] change views to class-based views from DRF


banesullivan:
The test is failing because `gdal_translate` is not in the current path in the tox test env when launched on CircleCI. I ssh'd into CircleCI and I can see `gdal_translate` under `~/project/.tox/test/bin/`, but that is not in the tox environment's path for some reason...

manthey:
I think the reason `gdal_translate` is failing is not because it can't be found, but because it is a small wrapper that redirects to the actual location of gdal_translate (deep in the GDAL python package), and therefore needs to be invoked with a shell (e.g., `subprocess.check_call(cmd, shell=True)`.

banesullivan:
Ah okay, I will try that.

FYI, it works when testing locally for me (within docker-compose), just not on CircleCI

banesullivan:
Same issue ^. When ssh'ing into CircleCI, I notice that the `bin` dir is not in the path which is what leads me to https://github.com/ResonantGeoData/ResonantGeoData/pull/209#issuecomment-736846881

```sh
circleci@d8ac92da9f01:~$ cd project/.tox/
circleci@d8ac92da9f01:~/project/.tox$ ls test/bin/gdal_*
test/bin/gdal_auth.py  test/bin/gdal_edit.py        test/bin/gdal_pansharpen.py  test/bin/gdal_retile.py
test/bin/gdal_calc.py  test/bin/gdal_fillnodata.py  test/bin/gdal_polygonize.py  test/bin/gdal_sieve.py
test/bin/gdal_contour  test/bin/gdal_grid           test/bin/gdal_proximity.py   test/bin/gdal_translate
test/bin/gdal_create   test/bin/gdal_merge.py       test/bin/gdal_rasterize      test/bin/gdal_viewshed
circleci@d8ac92da9f01:~/project/.tox$ which gdal_translate
circleci@d8ac92da9f01:~/project/.tox$ test/bin/python
Python 3.8.6 (default, Nov 18 2020, 13:49:49)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys, subprocess
>>> subprocess.check_call('echo $PATH', shell=True)
/home/circleci/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
0
>>> exit()
circleci@d8ac92da9f01:~/project/.tox$
```

So perhaps `subprocess.check_call(..., shell=True)` is executing a different shell environment than the one activated by tox where the `~/project/.tox/test/bin/` dir is supposed to be added to the path?

banesullivan:
I switched from using a subprocess to using gdal's Python bindings for `Translate` and all is working

banesullivan:
I am not going to add a test to verify that the produced COG is valid. IMO we don't need to test that as it is upstream functionality. We are passing the correct options to `gdal_translate` and we should accept what it returns, so the test currently just verifies that something is returned.

banesullivan:
Going to squash and merge. More endpoints/views are coming in a follow up PR for #201  and we can use that as a place to modify any of this work if needed

PR #212, opened 2020-12-02
Include static resources with installation

### LINKED ISSUE #210: Static resources different on deployment
It appears the static resources are set to something different when using the deployment settings. 

There is a deployment error from the stylesheet not being found. The dev settings work though.

Where is this set in the deployment configuration?

banesullivan:
Resolve #210 ?

I think this should do it. Thank for the pointer, @jbeezley

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,8 @@ setup(
     name='resonantgeodata',
     version='0.1',
     python_requires='>=3.8.0',
+    include_package_data=True,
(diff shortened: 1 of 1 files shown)


brianhelba (changes requested):
See my comment https://github.com/ResonantGeoData/ResonantGeoData/issues/210#issuecomment-744491587

I don't think this fixes the issue.

PR #214, opened 2020-12-14
Use field_file_to_local_path from upstream "django-girder-utils"

brianhelba:
This implementation is better maintained and supported across multiple projects.


brianhelba:
I'm going to be deleting the old `django-girders` package soon, so at a minimum, we need to remove that as a dependency.

Additionally, I'm hoping to increase adoption of `django-girder-utils` (the slightly-better-renaming), to relieve the burden of downstreams needing to maintain common utilities like this.

banesullivan (approved):

PR #217, opened 2020-12-15
Add new SubsampledImage model with endpoints

### LINKED ISSUE #201: Add endpoint to serve ROIs of images as raw arrays

banesullivan:
Resolve #201 

- New `SubsampledImage` with POST/GET endpoints.
  - POST endpoint for creating a subsampled image from the three options:
    - simple pixel coordinate bounding box: `umin,umax,vmin,vmax`
    - geospatial bounding box: `xmin,xmax,ymin,ymax` (world coordinates)
    - GeoJSON polygon we serve the region of the image within that polygon
  - Sample parameters are passed as JSON field
  - When using a GeoJSON polygon, a masked raster is returned and it cannot be in COG format.
- New wrapper for pairing with signal receivers: `skip_signal`
  - Makes it easy to set an attribute on an instance and have the event signal ignored. This is particularly useful to the `TaskEventMixin` where a celery task is launched on the `post_save` -- if we are calling save from within a task and need it to finish then we might want to synchronously call that code rather than launching another task.
- Improved admin interface for `ImageEntry` - now a thumbnail is shown in the list view
- Added GET endpoints to download files directly from the storage system (e.g. S3) for: `ArbitraryFile`, `ImageEntry`, and `ConvertedImageFile`
- Added new endpoint to GET a single `SpatialEntry` and fixed how the `outline` field is serialized
- Better error handling for the `download_file` endpoint
- Testing!
  - Basic tests to verify things are running as expected. These don't really verify the results/output


rgd/geodata/serializers.py
```diff
@@ -27,8 +29,55 @@ class Meta:
...
+class SubsampledImageSerializer(serializers.ModelSerializer):
+
+    data = serializers.HyperlinkedRelatedField(
+        many=False, read_only=True, view_name='arbitrary-file-data'
+    )
+
+    def to_representation(self, value):
+        ret = super().to_representation(value)
+        # TODO: fix domain
+        ret['status'] = reverse('subsampled-status', args=[value.id])
```
banesullivan, line 61: @mcovalt, any idea on how to resolve this URL to be a hyperlink? Currently, it outputs the API path without the host

banesullivan:
I added a notebook to show off these new endpoints in `demos/Geodata-API.ipynb`. It shows how to use a GeoJSON polygon to subsample a raster image entry and fetch the result:

| Source | Shape | Result |
|---|---|---|
| download |  download | download |

rgd/geodata/serializers.py
banesullivan: It gives:

```
'status': '/api/geodata/imagery/subsample/98/status',
```

but it needs to yield that path with the host domain, e.g.

```
'status': 'http://localhost:8000/api/geodata/imagery/subsample/98/status',
```

(or wherever it is hosted)

mcovalt, line 61: I'm working on getting sample data loaded. I think this will work though:

```suggestion
        realtive_status_uri = reverse('subsampled-status', args=[value.id])
        if 'request' in self.context:
            request = self.context['request']
            ret['status'] = request.build_absolute_uri(realtive_status_uri)
        else:
            ret['status'] = realtive_status_uri
```

mcovalt: Just tested and this works.

mcovalt: See https://github.com/ResonantGeoData/ResonantGeoData/pull/217/commits/cf33017ba9bb70734d88e690d8f6b0bbff9441e7.

mcovalt: See https://github.com/ResonantGeoData/ResonantGeoData/pull/217/commits/cf33017ba9bb70734d88e690d8f6b0bbff9441e7.

banesullivan: Awesome, thanks for figuring this out!

rgd/geodata/api/download.py
```diff
@@ -37,3 +37,72 @@ def download_file(request, model, id, field):
...
+
+
+@swagger_auto_schema(
+    method='GET',
+    operation_summary='Check the status of SubsampledImage.',
+)
+@api_view(['GET'])
+def get_status_subsampled_image(request, pk):
+    """Get the status of any SubsampledImage model."""
+    return _get_status_response(request, 'SubsampledImage', pk)
```
mcovalt, line 108: I'm not going to worry about translating this into a class based view right now. I think it would be difficult.

mcovalt (approved):
Looks good! I'll let you merge if you approve of my recent commit on the branch: https://github.com/ResonantGeoData/ResonantGeoData/pull/217/commits/cf33017ba9bb70734d88e690d8f6b0bbff9441e7.

rgd/geodata/tests/test_api.py
```diff
@@ -0,0 +1,136 @@
...
+            'sample_type': 'pixel box',
+            'sample_parameters': {'umax': 100, 'umin': 0, 'vmax': 200, 'vmin': 0},
+        },
+    )
+    # TODO: wait for the task kicked of by POST to complete?
+    sub = models.imagery.SubsampledImage.objects.get(source_image=elevation_image.id)
+    assert sub.data
+    # Test the GET
+    content = json.loads(client.get(f'/api/geodata/imagery/subsample/{sub.id}/').content)
+    assert content
```
banesullivan, line 119: This POST endpoint launches a celery task. How can we set up this test to wait on that task to complete?

Normally, the CELERY_EAGER flags capture the execution but in this case, since we are hitting the POST endpoint through the client it becomes async with the test execution.

banesullivan: @mcovalt or @manthey, any ideas on how to resolve this?

I'd like to test the endpoint itself, but I could change this to create a SubsampledImage model directly.

banesullivan: Nevermind, I figured this out. I was looking at the wrong issue and it turns out you have to use `json.dumps()` for any internal dicts in the payload.

PR #159, opened 2020-09-30
Add FMV test

banesullivan:
The test is failing because the docker client cannot launch from within the tox environment. I'm not sure how to resolve this after many attempts

Continuation of #143

The error (https://github.com/ResonantGeoData/ResonantGeoData/pull/143#issuecomment-693694449)

Happening at `client = docker.from_env(version='auto', timeout=3600)`

```
File "/opt/django/geodata/models/fmv/etl.py", line 25, in _extract_klv_with_docker
    client = docker.from_env(version='auto', timeout=3600)
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/client.py", line 84, in from_env
    return cls(
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/client.py", line 40, in __init__
    self.api = APIClient(*args, **kwargs)
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/api/client.py", line 188, in __init__
    self._version = self._retrieve_server_version()
  File "/opt/django/.tox/test/lib/python3.8/site-packages/docker/api/client.py", line 212, in _retrieve_server_version
    raise DockerException(
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

PR #221, opened 2020-12-29
Demo data population command

### LINKED ISSUE #160: database migration to initialize for demos
We should create a migration to populate the database with entries to be able to quickly spin up demos and have some sample data for folks to see on the deployed website.

It'd be great to aggregate example data (likely just what is included in the datastore for the `geodata` tests coming from data.kitware.com) and include those. We could also do the same for the `core` app.

I'm pretty sure this would be implemented in a manually created migration, is that the best way to go about it? 

Would everyone be on board with this, or should we have it as an optional feature maybe for the deployed site only?

banesullivan:
Resolve #160 

This adds a `manage.py` command to populate the database with demo data. Simply run:

```sh
./manage.py demo_data
```

I thought it would be easiest/best to have the Pooch registry keep track of all of these data as it is all stored on DKC. These data can be used for tests as needed. I have moved this out of the testing module and directly in geodata. I have also added tests to make sure all entries in the registry are valid and available for download.

I have added a fair amount of shapefiles from https://www.coloradoview.org/colorado-gis/ and I am working to add full coverage of CO with Landsat 8 soon as well.

Once we figure out the Docker issues in #159, we can add more FMV demo datasets

I would also like to add a full KWCOCO demo, but I want to refactor the BaseImageFile model to better handle COCO archives first... That will need to come in a later PR.

Also resolves a bug when loading shapefiles from an archive. Sometimes they might be in a nested directory upon extraction and this provides a way to handle that.

----

Inspiration: 
- https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/
- https://adamj.eu/tech/2020/09/07/how-to-unit-test-a-django-management-command/

Something else to look into: https://docs.djangoproject.com/en/3.1/ref/django-admin/#django-admin-loaddata

For the ManyToMany query, needed this: https://stackoverflow.com/questions/53911314/django-filter-by-many-to-many-with-exact-same-query/53911762#53911762


mcovalt:
Just tried this out. This is great!

rgd/geodata/tests/test_demo_data.py
```diff
@@ -0,0 +1,25 @@
...
+        "demo_data",
+        *args,
+        stdout=out,
+        stderr=StringIO(),
+        **kwargs,
+    )
+    return out.getvalue()
+
+
+@pytest.mark.skip
```
banesullivan, line 21: Marking test to skip because it takes ~107.25s with the current list of data (I plan to add more)

I'm going to keep the test here for manual running (that is, comment this line locally to make sure everything is working). It's just too much to put on the CIs and all other normal dev testing

banesullivan:
I have this set up so that the command does not re-add any data that was previously added. This means you can run the command infinitely many times without consequence and that to add more sample data, we simply add them to the registry and to the appropriate list in `demo_data.py`.

PR #230, opened 2021-01-08
Increase Heroku dyno sizes

brianhelba:


brianhelba:
If we decide this is too expensive, we can revert it later, but for now, it should help production to stop crashing.

PR #236, opened 2021-01-12
Prevent duplicate SubsampledImage models on POST

### LINKED ISSUE #229: SubsampledImage: prevent duplicate sample_parameters
We should do a check to see if the `sample_parameters` for any created `SubsampledImage` object match any existing entry to prevent duplication of data.

banesullivan:
Resolve #229 

The `SubsampledImageSerializer` now checks the DB for existing entries with the same sample parameters and returns the first match if one exists rather than creating a duplicate.

It also triggers a save event to reprocess the subsampling

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,6 +49,7 @@ passenv =
 deps =
     -rrequirements.txt
     factory-boy
+    faker<=5.4.0
     pytest
     pytest-django
     pytest-factoryboy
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -49,6 +49,7 @@ passenv =
 deps =
     -rrequirements.txt
     factory-boy
+    faker<=5.4.0
```
banesullivan, line 52: Issue in https://github.com/joke2k/faker/issues/1367

Limiting until they make a release to resolve this

PR #238, opened 2021-01-13
Use get_or_create in serializer

banesullivan:
Quick patch to make code cleaner/better and use functionality already exists in django

Follow up to #236 

-----

While I was at it, I cleaned up a few other places in the ETL routines where this logic was used and added a new `get_or_create_no_commit` helper for doing the same thing without automatically saving to the DB.

### DIFF
--- a/rgd/geodata/serializers.py
+++ b/rgd/geodata/serializers.py
@@ -84,23 +84,10 @@ class SubsampledImageSerializer(serializers.ModelSerializer):
         ]
 
     def create(self, validated_data):
-        """Prevent duplicated subsamples from being created.
-
-        This will check to see if an existing model matches the
-        ``sample_parameters`` for the given image ID and return that object
-        instead of creating another.
-
-        """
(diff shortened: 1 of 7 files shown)


rgd/geodata/serializers.py
```diff
@@ -84,23 +84,10 @@ class Meta:
...
-        q = models.SubsampledImage.objects.filter(
-            sample_parameters=sample_parameters, source_image=source_image
-        )
-        if q.count() == 0:
-            obj = models.SubsampledImage.objects.create(**validated_data)
-        else:
-            obj = q.first()
-            # Trigger a reprocessing
+        """Prevent duplicated subsamples from being created."""
+        obj, created = models.SubsampledImage.objects.get_or_create(**validated_data)
```
banesullivan, line 88: We could do `self.Meta.model.objects.get_or_create(**validated_data)` and let this be an abstract class we could reuse elsewhere. I'll keep that in mind if we need other serializers to have this functionality

PR #245, opened 2021-01-16
adjustments to SpatialEntryFilter

banesullivan:
Follow up to #208

I am performing manual testing of the `SpatialEntryFilter` to make sure everything is in working order.

I am hoping to add some basic tests for this as well

### DIFF
--- a/rgd/geodata/filters.py
+++ b/rgd/geodata/filters.py
@@ -127,7 +127,7 @@ class SpatialEntryFilter(filters.FilterSet):
 
     def filter_relates(self, queryset, name, value):
         """Filter the spatial entries by the chosen DE-9IM."""
-        if value != '':
+        if value:
             geom = self.form.cleaned_data['q']
             return queryset.filter(footprint__relates=(geom, value))
         return queryset
--- a/rgd/geodata/filters.py
+++ b/rgd/geodata/filters.py
@@ -152,7 +152,7 @@ class SpatialEntryFilter(filters.FilterSet):
         if value == 'geometry':
             return queryset.filter(geometryentry__isnull=False)
         if value == 'raster':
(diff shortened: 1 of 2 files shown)


mcovalt:
I've added a single simple test for rasters.

A full fledged test-suite would test each of...
* raster
* FMV
* geometry
 
...for each of the filter options. The result would be tons of tests! I'd like to help contribute more, but I'm curious where we should draw the testing line. Thoughts? Opinions?

rgd/geodata/filters.py
```diff
@@ -120,14 +120,14 @@ def filter_q(self, queryset, name, value):
 
     def filter_predicate(self, queryset, name, value):
         """Filter the spatial entries by the chosen predicate."""
-        if value is not None:
+        if value != '':
```
banesullivan, line 123:
```suggestion
        if value:
```

rgd/geodata/filters.py
```diff
@@ -120,14 +120,14 @@ def filter_q(self, queryset, name, value):
...
             geom = self.form.cleaned_data['q']
             return queryset.filter(**{f'footprint__{value}': geom})
         return queryset
 
     def filter_relates(self, queryset, name, value):
         """Filter the spatial entries by the chosen DE-9IM."""
-        if value is not None:
+        if value != '':
```
banesullivan, line 130:
```suggestion
        if value:
```

banesullivan:
I think that these basic tests are all we should worry about for now. Thanks, @mcovalt! In time, we can implement full coverage for the `filters.py` module but as long as we can verify the geometry/predicate work then I think that is sufficient.

I will add one additional test using GeoJSON and then merge for us to add more comprehensive tests later.

PR #248, opened 2021-01-22
Use authenticated API client

banesullivan:
Patch for https://github.com/girder/django-composed-configuration/pull/109

This changes the two tests hitting the POST endpoint to use an authenticated API client.

@brianhelba: for endpoints that are read-only, should we be using [`api_client` here](https://github.com/ResonantGeoData/ResonantGeoData/blob/87ba08fd261c34940d615c1f0e13874dc333e487/rgd/geodata/tests/conftest.py#L12) over the default `client` fixture (presumably coming from `pytest-django`... or where is that actually coming from?)?

### DIFF
--- a/rgd/geodata/tests/test_api.py
+++ b/rgd/geodata/tests/test_api.py
@@ -52,48 +52,48 @@ def landsat_raster():
 
 
 @pytest.mark.django_db(transaction=True)
-def test_download_file(client, arbitrary_file):
+def test_download_file(api_client, arbitrary_file):
     model = 'ArbitraryFile'
     id = arbitrary_file.id
     field = 'file'
-    response = client.get(f'/api/geodata/download/{model}/{id}/{field}')
+    response = api_client.get(f'/api/geodata/download/{model}/{id}/{field}')
     assert response.status_code == 200, response.content
     with pytest.raises(AttributeError):
         # Test bad model
-        client.get('/api/geodata/download/Foo/0/file')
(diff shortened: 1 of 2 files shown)


brianhelba:
@banesullivan Great question. Of the two fixtures:
* `client` is [defined by `pytest-django`](https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client). As you see, it provides an instance of [`django.test.Client`, which is part of Django proper and documented there](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#the-test-client). It's designed around making requests to server-rendered pages, since that's what Django proper is designed for.
* `api_client` is [defined by ResonantGeoData (created by the Girder 4 cookiecutter)](https://github.com/ResonantGeoData/ResonantGeoData/blob/87ba08fd261c34940d615c1f0e13874dc333e487/rgd/geodata/tests/conftest.py#L12). It provides an instance of [`rest_framework.test.APIClient`, which is part of Django Rest Framework and documented there](https://www.django-rest-framework.org/api-guide/testing/#apiclient). It further extends the basic Django test client and is designed around making RESTful requests to DRF during testing.

So, if you're testing a server-rendered page, use `client`. If you're testing a DRF-provided REST API endpoint, use `api_client`.

brianhelba (approved):

banesullivan:
Ah! Thank you for clarifying this!

Since all of our tests are currently hitting DRF, I will make the appropriate switch as a part of this PR

PR #251, opened 2021-01-25
Improve examination of HTTP response data in testing

brianhelba:
This improves 2 things:
* Use `response.data` instead of `json.loads(response.content)`. This still ensures that the DRF Serializers are invoked to serialize Model representations into primitive Python data structures, but avoids the code and performance overhead of encoding / decoding a JSON string. See: https://www.django-rest-framework.org/api-guide/testing/#checking-the-response-data (see #252)
* Perform each assert on its own line, which is syntactically necessary. Additional expressions to an `assert` statement are not interprepted as additional assertions to be made, but rather as the arguments (such as a failure message) to the `AssertionError` that a failure raises. See: https://docs.python.org/3/reference/simple_stmts.html#grammar-token-assert-stmt

### DIFF
--- a/rgd/core/urls.py
+++ b/rgd/core/urls.py
@@ -50,7 +50,6 @@ urlpatterns = [
     path('jobs/<str:creator>/<int:pk>/', views.JobDetailView.as_view(), name='job-detail'),
     path('tasks/', views.tasks, name='tasks'),
     path('task/<int:pk>-<str:name>/', views.TaskDetailView.as_view(), name='task-detail'),
-    path('api/core/download/<model>/<int:id>/<field>', views.download_file, name='download-file'),
     path('', include(router.urls)),
 ] + generate_routes({'flower-proxy': {'base_url': 'http://flower:5555/', 'prefix': '/flower/'}})
 
--- a/rgd/core/views.py
+++ b/rgd/core/views.py
@@ -1,19 +1,12 @@
 import logging
-import os
 
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
-from django.db.models.fields.files import FieldFile
-from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, render
 from django.utils import timezone
-from django.utils.encoding import smart_str
 from django.views.decorators.http import require_http_methods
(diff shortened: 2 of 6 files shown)


brianhelba:
@banesullivan @mcovalt 
Also note, there are a lot of places where multiple separable test cases are grouped under a single function. While some workflows may require a sequential series of operations to strictly occur in order within a single test function, anything else that can be separated into multiple functions and run / asserted independently should be. This has a few benefits:
* It allows failures to be more clearly diagnosed, since function names can be more specific to the semantics of the test case within, and since one failing case in a function doesn't block others from running.
* It prevents errors due to accidental shared state within a test function.
* It allows tests to be sped up in the future by parallelizing their execution.

Another useful rule of thumb is that each test function, particularly for unit tests, should ideally follow the 3-stage stricture:
* Setup pre-conditions
  * Ideally, this can often be done entirely via fixtures.
* Execute the logic to be tested
* Assert a single fact
  * Depending on the data structure, this might need to be made with multiple `assert` statements, but it should all be asserting only a single logical aspect per case

Of course, this is aspirational, and practical tests may need to bend the rules to be manageable, but I think it makes the tests much more maintainable in the long-term.

banesullivan:
> Perform each assert on its own line, which is syntactically necessary. Additional expressions to an assert statement are not interprepted as additional assertions to be made, but rather as the arguments (such as a failure message) to the AssertionError that a failure raises. See: https://docs.python.org/3/reference/simple_stmts.html#grammar-token-assert-stmt

@brianhelba, this wasn't intended to assert that the returned data isn't `None`. I was using `.content` as the message to return (this was helpful during debugging to see what the endpoint was returning when not yielding the correct status code). I'll admit this was an artifact of debugging during development and I am in favor of removing them.

Unfortunately, we still use `HttpResponseRedirect` in some of the endpoints to point to the data URL of files coming from S3-file-field. These responses do not have `.data`... should we change those redirects to use `rest_framework.response.Response` as well?

banesullivan:
On your note about testing practices: I agree entirely and was thinking about this recently. Most of the tests were written the way they are for time's sake and I will work to create more separable tests with a new `fixtures.py` module we can re-use. That will have to come later, but I will definitely address these concerns.

banesullivan (approved):

PR #252, opened 2021-01-25
Return DRF Response objects in API endpoints

brianhelba:
DRF's "Response" class handles the details of content negotation and JSON encoding. It should almost always be used as the return value of API endpoints. See: https://www.django-rest-framework.org/api-guide/responses/


banesullivan (approved):
Thanks for catching this, @brianhelba. I will need to make sure to do this moving forward

PR #255, opened 2021-01-28
Check that content yeilds from download endpoints

banesullivan:
I cannot figure out why these tests are failing. The response from these endpoints has content when I hit them with Swagger or the demo python notebook. For some reason, they aren't yielding anything in the tests only.

Here is the `api/geodata/imagery/image_entry/{pk}/data` endpoint in swagger with response content 😕 

Screen Shot 2021-01-28 at 10 00 38 AM

Yet the test for that endpoint fails with: 

```
@pytest.mark.django_db(transaction=True)
    def test_download_image_entry_file(api_client, astro_image):
        pk = astro_image.pk
        response = api_client.get(f'/api/geodata/imagery/image_entry/{pk}/data')
        assert status.is_redirect(response.status_code)
>       assert response.content
E       assert b''
E        +  where b'' = <HttpResponseRedirect status_code=302, "text/html; charset=utf-8", url="http://localhost:9000/test-django-storage/3f5f...ires=604800&X-Amz-SignedHeaders=host&X-Amz-Signature=201e4ffdb18f6dda36dffdb5eee8450184ce9ea56967339c063ccc0049fa3a4c">.content

api_client = <rest_framework.test.APIClient object at 0x7f10c46df9d0>
astro_image = <ImageEntry: astro.png (12)>
pk         = 12
response   = <HttpResponseRedirect status_code=302, "text/html; charset=utf-8", url="http://localhost:9000/test-django-storage/3f5f...ires=604800&X-Amz-SignedHeaders=host&X-Amz-Signature=201e4ffdb18f6dda36dffdb5eee8450184ce9ea56967339c063ccc0049fa3a4c">
```


banesullivan:
@zachmullen, would you be able to look at this and/or decide if this is an issue we should spend time on

zachmullen:
The download endpoint responds with a HTTP 302 redirect. It has no body, but does have a `Location` response header that redirects to minio or S3. Following that redirect under normal 302 behavior will then download the bytes in the body of that request.

If you want to assert the content is actually being downloaded, you can either set `follow=True` on the `api_client.get` call, or manually follow it with a follow-up request and then you can make assertions about the whole flow.

banesullivan:
Ah, thanks for the insight!!

> If you want to assert the content is actually being downloaded, you can either set follow=True on the api_client.get call, or manually follow it with a follow-up request and then you can make assertions about the whole flow.

I will set up a follow-up request to assert that content is actually being downloaded. I think it would be good to assert that the redirect actually points to a download link

PR #257, opened 2021-01-29
FUSE file URLs to avoid full downloads

### LINKED ISSUE #237: Allow lazy random access to files in Storage on S3
When a Django `FieldFile` (the object referenced by a `FileField`) is opened, both in development with MinIO and in production with AWS S3, the `Storage` layers download a whole file write it to a `SpooledTemporaryFile` (which is in-memory for small content and on-disk for larger content)
* `S3Boto3Storage`: https://github.com/jschneier/django-storages/blob/770332b598712da27ecdba75c9e202ad6a1a8722/storages/backends/s3boto3.py#L132
* `MinioStorage`: https://github.com/py-pa/django-minio-storage/blob/8ce77cd8eae0d86fa3200e74cde64144c0338f37/minio_storage/files.py#L130

However, many operations need random access to only a small part of the file's content.

If all that's necessary is a Python file-like object (which is the same interface that a Django `FieldFile` naturally provides), we could look into developing a new `Storage` wrapper which lazily loads the content of the file-like object, internally using HTTP Range requests to only download the required parts just-in-time to be read from. This would likely also include some caching of the parts once they are retrieved.

If external tools (like GDAL) need access via a filesystem path (i.e. what's provided by `field_file_to_local_path`), then FUSE may be the most appropriate abstraction. However, FUSE could not be run on Heroku and might generally be more difficult to install for new developers.

In either case (Python file-like object or FUSE), the underlying logic for how to fetch and cache portions of file content may be the same.

banesullivan:
When combined with #276, this resolves #237 

This has the potential to dramatically improve data ingestion times (from minutes to single-digit seconds for most rasters). I am still working on this to find all the bottlenecks.

See https://github.com/ResonantGeoData/ResonantGeoData/issues/237#issuecomment-769291382 for details. This only works when run through `docker-compose`... since we are currently launching the worker on my local machine through docker-compose, this will work. However, it will not work on a Heroku worker. In the future, we will have to set up an EC2 worker.

This expands on #272 to improve ETL routines such that it is optimized to quickly ingest data and later refine that data

### Todo

- [x] Fix raster ETL routine `_reproject_raster` to subsample before reprojecting
- [x] Update `landsat_data` command to ingest more data now that it is inexpensive
- [ ] Will this work for FMV ETL routines where the file is shared with a Docker container?

### Limitations

- Cannot reference other files/relative paths (see https://github.com/ResonantGeoData/ResonantGeoData/issues/237#issuecomment-769318990 and #273 )
- The entire file is still downloaded for computing the hash/checksum for the `ChecksumFile` model
- Cannot add a FUSE bypass to `field_file_to_local_path` for S3FileField


banesullivan:
I'm still having an issue when creating a thumbnail/subsampled raster where the total runtime is so long that it seems like the entire file is being downloaded (similar execution times). Perhaps `simple-httpfs` doesn't allow random access to remote files but rather sequential? Some further investigation is needed...

PR #264, opened 2021-02-04
Expose facet search in UI

banesullivan:
Exposes the `SpatialEntryFilter` for the search pages in the current front-end.

This requires us to bump to Django 3.2

Todo:

- [x] Remove specific data type listing pages (use the new datatype filter parameter to choose type)
- [x] Add DateTime fields for the filter
- [x] Fix the distance fields in form
- [x] UI: Label min/max fields in the search form
- [x] UI: add a clear button to datetime fields
- [x] UI: improve list view of search results
- [x] Figure out how to map the detail page link properly for each data type


setup.py
```diff
@@ -37,7 +37,7 @@
     include_package_data=True,
     install_requires=[
         'celery',
-        'django',
+        'django==3.2a1',
```
banesullivan, line 40: Due to this bug: https://code.djangoproject.com/ticket/31910

brianhelba: It'd be good to note this as a code comment. That should be more persistent.

banesullivan: Will do

PR #268, opened 2021-02-12
Upgrade to cookiecutter release v0.17

brianhelba:
This supports the additional changes:
* Use `PIP_FIND_LINKS` environment var instead of `requirements.txt`
  * Any pip options can also be set via environment variables. With this change, RGD now has a single source of truth for its dependencies (`setup.py`).
* Remove `_account.scss`, as `django-girder-style` now provides this styling
* Add missing `Pillow` requirement


brianhelba:
@banesullivan I'm not clear on the actual purpose of the `worker` extra. In particular, under what circumstances should it **not** get installed?

tox.ini
```diff
@@ -41,14 +41,17 @@ commands =
...
     DJANGO_DATABASE_URL
     DJANGO_MINIO_STORAGE_ACCESS_KEY
     DJANGO_MINIO_STORAGE_ENDPOINT
     DJANGO_MINIO_STORAGE_SECRET_KEY
+extras =
+    dev
 deps =
-    -rrequirements.txt
     factory-boy
     faker<=5.4.0
```
brianhelba, line 56: I'm also curious, why was `faker` pinned?

banesullivan: I ran into a bug after one of their releases. See https://github.com/ResonantGeoData/ResonantGeoData/pull/236#discussion_r555914603

banesullivan:
> @banesullivan I'm not clear on the actual purpose of the `worker` extra. In particular, under what circumstances should it **not** get installed?

My intention was to prevent some of the heavyweight depends from being installed in the Heroku server instance and just keep those packages in the worker. Many our depends are only used in the ETL routines (triggered by celery tasks) so they do not need to be installed/running within the app

brianhelba:
> My intention was to prevent some of the heavyweight depends from being installed in the Heroku server instance

The `worker` extras are currently getting installed on Heroku: https://github.com/ResonantGeoData/ResonantGeoData/blob/8a4fdcb10f4d4bc445d33564372a83f1351f9ee2/requirements.txt#L5

I'm not sure if there's a way to install a separate set of dependencies on a per-dyno basis. On the other hand, we're not using Heroku to run workers.

banesullivan:
I'm all for removing the `[worker]` flag in `requirements.txt` to clean up the heroku server since we're currently running the worker on my local machine via doker-compose which installs the worker deps:


https://github.com/ResonantGeoData/ResonantGeoData/blob/8a4fdcb10f4d4bc445d33564372a83f1351f9ee2/dev/django.Dockerfile#L30

banesullivan (approved):

PR #276, opened 2021-02-28
Large image

### LINKED ISSUE #260: Large image application
A small part of https://github.com/girder/large_image/issues/524 needs to be implemented for this Django project. The parts needed are:

- [ ] endpoint or method to make / unmake a Django file field into a large_image item
- [ ] endpoints for metadata (/tiles)
- [ ] endpoints for serving tiles (/tiles/zxy)

mcovalt:
Closes #260

This is a simple tile server endpoint. It's not an app as was imagined. There's a 💯abstract™ app in 701293890f75b646d042da877289646ebac88a7f. It started feeling silly for just one endpoint, so it was reverted. It may be useful down the road.

I didn't want to deal with the context manager to download/cleanup S3 files, so instead I'm using [`/vsis3/`](https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files) and [`/vsicurl/`](https://gdal.org/user/virtual_file_systems.html#vsicurl-http-https-ftp-files-random-access). You can see the implementation in 8b4d2fc6f162447c128b1a4fdb17e8353853dd09.  I think it's simpler and luckily results in a pretty quick tile server. It was tested with other ETL routines in ed9bd8ae57e3d9c9ff9f891847a051073280b2ce, which seemed to work fine, but that was reverted to stay on topic with this PR.

New endpoints:
- `api/geodata/imagery/image_entry/<pk>/tiles`: get an `ImageEntry`'s metadata
- `api/geodata/imagery/image_entry/<pk>/tiles/<z>/<x>/<y>.jpg`: get an `ImageEntry`'s `x, y` tile at zoom-level `z`


manthey:
Using `vsis3` is probably sufficient for RGD at this time, but ultimately won't be sufficient for the various other use cases we have for large_image.  Moreover, since `vsis3` relies on some environment variables for the S3 access setup, it restricts us to have a single assetstore.  For instance, in Girder 3 deployments, we can use multiple s3 buckets with completely different options to have data from multiple sources.

In other projects we've had a small set of geospatial data in formats GDAL doesn't read (such as some netcdf files) that we were able to use the mapnik tilesource to handle.  If we want to handle such datasets, we'll either need to have an appropriate ETL step or go back to something like fuse to provide mapnik access to such files.

mcovalt:
Ah, shoot. I was hoping we could get away with avoiding FUSE since it can make deployment more difficult. Does mapnick have its own API to access files over HTTP? I could see maintaining a bunch of different methods for remote file access being a burden, though—more so than FUSE.

> `vsis3` relies on some environment variables for the S3 access setup, it restricts us to have a single assetstore

I was thinking `vsicurl` could be a work-around. It seems like it is capable of a per-request setup that avoids the global state needed for S3. The only benefits of `vsis3` from what I can tell is that it also allows writing to S3 nicely.

manthey:
I don't think mapnik has any api access for http; it relies on GDAL where it can.  I think what you have here is sufficient for RGD until we need to handle associated files or have files that we need to read with mapnik.

As a general solution for large_image and Django, fuse is attractive in that more tile sources don't have any way to do http access on their own (libtiff, openslide, nd2, openjpeg, etc) and that we can eventually write or find an implementation where files that are relative to each other are exposed properly.

mcovalt:
@banesullivan In #260, you mentioned the endpoint could accept a URL and produce a result from that. Instead, this endpoint accepts an `ImageEntry.pk`. Is this OK? I did this to simplify handling permissions.

banesullivan:
> this endpoint accepts an ImageEntry.pk. Is this OK? I did this to simplify handling permissions.

This should be okay. Under the hood, you are essentially still doing what I was getting at with `get_vsi_path()`. `large_image` accepts a URL/vsi path regardless of whether the file was added through S3FileField or a URL address and you just route that through the model first for permissions. 

I'm happy with this approach.... though down the road I may want to see our choice of remote file access (e.g. vsi vs FUSE) be standardized throughout the app, but that can come later as being discussed in #237

rgd/settings.py
```diff
@@ -13,6 +13,26 @@
...
 
 
+class GdalMinioMixin(ConfigMixin):
+    """Setup S3-like environment variables for Minio setups."""
+
+    @staticmethod
+    def before_binding(configuration: Type[ComposedConfiguration]):
+        # `MINIO_STORAGE_ACCESS_KEY` and `MINIO_STORAGE_SECRET_KEY`
+        # can only be accessed via environment variables
+        os.environ.setdefault(
```
brianhelba, line 23: It's kind of an abuse to use this function to do anything other than patch up dynamic settings. If you want to do something when the app starts up, I'd suggest this: https://docs.djangoproject.com/en/3.1/ref/applications/#django.apps.AppConfig.ready

rgd/settings.py
```diff
@@ -13,6 +13,26 @@
...
+            'AWS_ACCESS_KEY_ID',
+            os.environ['DJANGO_MINIO_STORAGE_ACCESS_KEY'],
```
brianhelba, line 25: To be stable with `composed-configuration`, these should be accessed via settings, not directly via the environment.

brianhelba:
Is this intended to be used only during development? How would this work during production, where AWS S3 is used instead of MinIO?

mcovalt:
> Is this intended to be used only during development?

This is intended to be used in both development and production. GDAL and Rasterio have different ways to configure S3 access. However, they can both be configured using the environment variables loaded in by [`GdalMinioMixin`](https://github.com/ResonantGeoData/ResonantGeoData/blob/777ac184abadbebf4bc7a863f28b10fb4072b4aa/rgd/settings.py#L16-L33).

> How would this work during production, where AWS S3 is used instead of MinIO?

From what I can tell in [django-composed-configuration](https://github.com/girder/django-composed-configuration/blob/e65bdc513d74ead19035a837079dad6df5ddf48c/composed_configuration/_storage.py#L54-L67), these environment variables are already set in production.

There exists a way to globally set up GDAL's use of S3 without using environment variables, but I could not see a way to do something similar with Rasterio. I didn't want to include dev-enviornment-only code where we use Rasterio, so setting up environment variables in the configuration mixins seemed to be a compromise.

rgd/settings.py
mcovalt: Both `configuration.MINIO_STORAGE_ACCESS_KEY` and `configuration.MINIO_STORAGE_SECRET_KEY` seems to be a `configurations.values.SecretValue` with a value of `None`. Any idea how to work around this?

rgd/settings.py
mcovalt: We only want this behavior when using [django-composed-configuration's `MinioStorageMixin`](https://github.com/girder/django-composed-configuration/blob/e65bdc513d74ead19035a837079dad6df5ddf48c/composed_configuration/_storage.py#L14). I am super uninformed on best practice regarding `ComposedConfiguration`. Not sure if `AppConfig.ready` seems right since this behavior shouldn't be run on production. Thoughts?

banesullivan:
I would also like to add a few things to this PR (I can help with these)

- [ ] An endpoint for fetching a thumbnail. I beleive it just needs to hook up to: [`getThumbnail()`](http://girder.github.io/large_image/large_image_source_gdal/large_image_source_gdal.html#large_image_source_gdal.GDALFileTileSource.getThumbnail): `api/geodata/imagery/image_entry/<pk>/thumbnail`
- [ ] Remove `Thumbnail` model and associated ETL routines (I can handle this)
- [ ] Hook up the raster detail view to load the image entry as a tile source with GeoJS (in `rgd/geodata/templates/geodata/rastermetaentry_detail.html`)
  - This will require removing the current quad and thumbnail that are shown in favor of the following:
  ```js
  var tiles = map.createLayer('osm', {
      url: 'http://localhost:8000/api/geodata/imagery/image_entry/{{ image.pk }}/tiles/{z}/{x}/{y}.jpg',
    });
  ```
  - I can handle this as well as I see a clear way to quickly do this and still hook it up to the dropdown interface I made for choosing which image in the raster to show
- [ ] Edit existing endpoints for raster/image details to return the tiles endpoint in the body for quick access

banesullivan:
All that is left to do is figure out why the MINIO storage URLs cannot be accessed from within the application container. I'm seeing 403s on the file URLs

banesullivan (approved):

rgd/utility.py
```diff
@@ -7,12 +7,14 @@
...
 from urllib.request import urlopen
 
+from django.conf import settings
 from django.db.models import fields
 from django.db.models.fields import AutoField
 from django.db.models.fields.files import FieldFile, FileField
 from django.http import QueryDict
 from django.utils.safestring import mark_safe
 from django_filters.rest_framework import DjangoFilterBackend
+from minio_storage.storage import MinioStorage
```
brianhelba, line 17: @mcovalt This will not work in production, since `minio_storage` is not installed there.

mcovalt: Fixed in 38d8e7af114fc4ef39656f369b9195c3f0d6d685

PR #288, opened 2021-03-04
Hotfix: gracefully handle when 'minio_storage' is not installed

mcovalt:
See https://github.com/ResonantGeoData/ResonantGeoData/pull/276#discussion_r587909564

### DIFF
--- a/rgd/utility.py
+++ b/rgd/utility.py
@@ -14,8 +14,12 @@ from django.db.models.fields.files import FieldFile, FileField
 from django.http import QueryDict
 from django.utils.safestring import mark_safe
 from django_filters.rest_framework import DjangoFilterBackend
-from minio_storage.storage import MinioStorage
 from rest_framework import parsers, serializers, viewsets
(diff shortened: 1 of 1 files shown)


rgd/utility.py
```diff
@@ -183,7 +187,8 @@ def patch_internal_presign(f: FieldFile):
     odd situation of accessing the file locally.
     """
     if (
-        isinstance(f.storage, MinioStorage)
+        MinioStorage is not None
```
brianhelba, line 190: You really don't need this check at all. If `MINIO_STORAGE_MEDIA_URL` is defined, I think it's safe to assume that the storage provider is `MinioStorage `, particularly since the `MINIO_STORAGE_MEDIA_URL` setting is not defined (and thus can never be picked up even if the environment variable is set) in `ProductionConfig`: https://github.com/girder/django-composed-configuration/blob/master/composed_configuration/_configuration.py#L58

PR #289, opened 2021-03-05
[CI] Move linting to its own step

banesullivan:
This moves the tests and linting into their own steps on CircleCI so it's easier to see what's going on/why a build failed

| `master` | this branch |
|---|---|
| Screen Shot 2021-03-05 at 12 08 08 PM | Screen Shot 2021-03-05 at 12 07 31 PM |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,14 @@ jobs:
           name: Run linting
           command: tox -e lint
       - run:
+          name: Run check migrations
+          command: tox -e check-migrations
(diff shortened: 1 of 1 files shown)


brianhelba:
What about: https://github.com/ResonantGeoData/ResonantGeoData/blob/38d8e7af114fc4ef39656f369b9195c3f0d6d685/tox.ini#L5 ?

banesullivan:
Missed that, whoops. I will update to put each in their own step

PR #292, opened 2021-03-08
Move patch_internal_presign

banesullivan:
@mcovalt, is it okay to move `patch_internal_presign` into `get_url` directly? Sometimes we need the minio URL if doing FUSE and not VFS

Also I renamed `get_local_vsi_path()` to `get_vsi_path()` as this should work in any environment

### DIFF
--- a/rgd/geodata/models/common.py
+++ b/rgd/geodata/models/common.py
@@ -223,9 +223,17 @@ class ChecksumFile(ModifiableEntry, TaskEventMixin):
             return url_file_to_local_path(self.url)
 
     def get_url(self):
-        """Get the URL of the stored resource."""
+        """Get the URL of the stored resource.
+
+        In most cases this URL will be accessible from anywhere. In some cases,
+        this URL will only be accessible from within the container. See
+        `patch_internal_presign` for more details.
+
+        """
         if self.type == FileSourceType.FILE_FIELD:
-            return self.file.url
+            with patch_internal_presign(self.file):
+                url = self.file.url
+            return url
         elif self.type == FileSourceType.URL:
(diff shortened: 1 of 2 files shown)


mcovalt:
I'm not sure. It should be OK if the URL returned by `ChecksumFile.get_url()` will _always_ return a URL to be used within the container. Sometimes the URL is to be used externally, like in the browser, though.

banesullivan:
> will always return a URL to be used within the container

Ah, shoot. Yep, I just noticed this breaks the `DATA LINK` column in the `ChecksumFile` admin list view that uses this to fetch the file URL.

Perhaps I add a flag to `get_url` such as `internal=False` that when `True` will use `patch_internal_presign`. Then have all of the internal code (ETL routines and tiler) set that flag?

rgd/geodata/models/common.py
```diff
@@ -206,26 +206,42 @@ def yield_local_path(self, vsi=False):
...
             is being utilized by GDAL and FUSE is not set up.
 
         """
         if self.type == FileSourceType.URL and precheck_fuse(self.get_url()):
-            return url_file_to_fuse_path(self.get_url())
+            return url_file_to_fuse_path(self.get_url(internal=True))
         elif vsi:
             logger.info('`yield_local_path` falling back to Virtual File System URL.')
-            return self.yield_local_vsi_path()
+            return self.yield_vsi_path(internal=True)
```
banesullivan, line 217: Hardsetting `internal=True` here because `yield_local_path()` should only ever be used in a local context for ETL routines or other server-side operations

mcovalt (approved):

PR #296, opened 2021-03-09
Demo data command improvements

banesullivan:
This improves the Landsat 8 demo data command to include acquisition dates when populating the database.

The command (use the `-c` argument to control how many rasters to ingest).

```
docker-compose run --rm django ./manage.py landsat_data -c 10
```

Note that I have this set up to only pull the RGB bands of the landsat imagery.

cc @zachmullen


banesullivan:
Note that I have the celery tasks run after all the initial entries are created because of a flaky bug that I will outline in a new issue

rgd/geodata/management/commands/_data_helper.py
```diff
@@ -63,9 +63,20 @@ def load_image_files(image_files):
...
+def load_raster_files(raster_files, dates=None, names=None):
+
+    if isinstance(raster_files, dict):
+        files = []
+        dates = []
+        names = []
+        for name, rf in raster_files.items():
+            files.append([rf['R'], rf['G'], rf['B']])
+            dates.append(rf['acquisition'])
+            names.append(name)
```
banesullivan, line 75: TODO: handle cloud cover after #310

PR #298, opened 2021-03-10
Cleanup TaskEventMixin

zachmullen:
* Make it an abstract model mixin so that `status` can reside inside it
* Add type annotations
* Remove superfluous runtime type validations (they were not sufficient anyway)


rgd/geodata/models/mixins.py
```diff
@@ -22,16 +23,16 @@ class TaskEventMixin(object):
...
-        if not isinstance(self.task_funcs, Iterable):
-            self.task_funcs = (self.task_funcs,)
-        if len(self.task_funcs) < 1:
+    status = models.CharField(max_length=20, default=Status.CREATED, choices=Status.choices)
+    task_funcs: List[Task] = []
+
+    def _run_tasks(self) -> None:
+        if not self.task_funcs:
             return
-        if not all([callable(f) for f in self.task_funcs]):
```
zachmullen, line 33: FYI, as a matter of good code habits for whoever originally wrote this: `any/all([...])` should just be `any/all(...)` to allow for short-circuiting.

banesullivan (commented):
We should also move `failure_reason` to the `TaskEventMixin` model now that it isn't abstract

rgd/geodata/models/mixins.py
banesullivan: Ah, thanks for the tip!

zachmullen:
Looks like I'm not authorized to merge to master

Issue #297, opened 2021-03-09
psycopg2.errors.UniqueViolation: duplicate key value error


banesullivan:
When running the demo data commands that I have, if the celery worker is set up to run in the background, an integretiy error for duplicate keys happens on the `image_entry.save()` call here:

https://github.com/ResonantGeoData/ResonantGeoData/blob/998a6c3995b4421c3632979a249fb78d66e1108f/rgd/geodata/models/imagery/etl.py#L69

The error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "geodata_imageentry_image_file_id_key"
DETAIL:  Key (image_file_id)=(14) already exists.
```

This is making me think that when we create a new `ImageEntry` in the tasks, there is some sort of race condition between jobs for the same `ImageFile`... which shouldn't happen? I'm not really sure what is going on here.
 

## Steps to reproduce

1. Clear the database volume
2. Apply migrations: `docker-compose run --rm django ./manage.py migrate`
3. In one session, launch the celery worker: `docker-compose up celery` and wait until ready
4. In another session, run the Landsat demo data command: `docker-compose run --rm django ./manage.py landsat_data -c 3`
    - Use the changes from #296 
5. Observe the error

## Error Message

<details>

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "geodata_imageentry_image_file_id_key"
DETAIL:  Key (image_file_id)=(14) already exists.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./manage.py", line 28, in <module>
    main()
  File "./manage.py", line 24, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
[...]

PR #299, opened 2021-03-10
Fix skip_signal typo

banesullivan:
Quick patch for #297 

This resolves the issue, but I want to keep #297 open until we have a better way to handle the `skip_signal` functionality to prevent typos like this from happening again


zachmullen (approved):
Let's merge this to get a fix deployed. Long term I think the best solution would be to make our task launching explicit.

PR #303, opened 2021-03-10
Use django_extensions TimeStampedModel

zachmullen:
Note that this requires a full database reset due to blowing away the old migrations. There would have been quite a lot of logic necessary in migration scripts to accommodate this change otherwise.


zachmullen:
One unexpected consequence of this change was that it revealed some (presumably) cargo cult stuff happening with relational references. A lot of places had an `ImageFile` object where we had referenced an `imagefile` attribute on it, which is just a pointer back to itself. That only worked because of the inheritance model that was previously being used.

banesullivan (approved):

rgd/geodata/api/search.py
```diff
@@ -420,8 +421,12 @@ def extent_summary(found, has_created=False):
...
         if has_created:
             summary = found.aggregate(
-                acquisition__min=Min(Coalesce('acquisition_date', 'created')),
-                acquisition__max=Max(Coalesce('acquisition_date', 'created')),
+                acquisition__min=Min(
+                    Coalesce('acquisition_date', 'created'), output_field=DateTimeField()
+                ),
+                acquisition__max=Max(
+                    Coalesce('acquisition_date', 'created'), output_field=DateTimeField()
+                ),
```
mcovalt, line 429: I have a brilliantly worded comment in [`SpatialEntryFilter.filter_acquired`](https://github.com/ResonantGeoData/ResonantGeoData/blob/timestamped-model/rgd/geodata/filters.py#L165-L166) regarding this bit:

> There is a special where filtering by acquisition, created is also filtered for `RasterEntry`.

The clarity is haunting. I think I remember noticing that `has_created` is only passed for rasters and replicated that logic over on the filter.

What's actually happening? What is the "special" for `RasterEntry`?

banesullivan:
Please hold until after [`v0.1.0`](https://github.com/ResonantGeoData/ResonantGeoData/milestone/1)

rgd/geodata/api/search.py
mcovalt: If https://github.com/ResonantGeoData/ResonantGeoData/pull/356 gets merged we can drop this.

PR #306, opened 2021-03-10
Improve docker-compose and FMV ETL

### LINKED ISSUE #285: docker-compose updates
- [ ] Add a second worker service and task queue
  - Have two queues: one for running fast, high-priority ETL routines (e.g. `populate_raster_entry`) and one for running slow, low-priority ETL routines (e.g. `populate_raster_footprint`)
  - Note changes in #294 
- [x] Split `django` and `celery` to use separate images to minimize unused dependencies in the app instance
- [x] Build the `celery` image against the `kitware/kwiver` image to avoid calling Docker from FMV ETL routine
- [x] Make sure we have a way of running the app in a non-privileged way after #257 
  - Somehow make `libfuse` and `simple_httpfs` optional in the Dockerfile

banesullivan:
Work to address #285 

- Adds `kwiver` wheel dependency to `celery` docker image. Ref https://github.com/Kitware/kwiver/issues/1239. This is an optional dependency in `setup.py`
- Moves `simple-httpfs` to optional dependency in `setup.py`. Only the celery worker image has FUSE installed now. This can be removed if we need the worker to run in a non-privileged mode.
- Adds new test for FMV ETL routine that can run on the CI
- adds a `.dockerignore` to reduce build context
- Updates detail view endpoint URLs (this has been bugging me for a while so I just did it here)

-----

New image sizes:

```
REPOSITORY                      TAG               IMAGE ID       CREATED          SIZE
resonantgeodata_django         latest            282ac4ac27ad   20 minutes ago   959MB
resonantgeodata_celery         latest            1d92cd6b2506   22 minutes ago   2.31GB
```


tox.ini
```diff
@@ -52,6 +52,7 @@ passenv =
 extras =
     dev
     worker
+    fmv
```
banesullivan, line 55: `kwiver` only has manylinux wheels which means we cannot run this natively on mac or windows. 

@ResonantGeoData/developers, if the tests break for you, this is likely why.

banesullivan: And this will require ffmpeg

rgd/geodata/urls.py
```diff
@@ -16,17 +16,17 @@
...
     ),
     path(
-        'geodata/fmv_entries/<int:pk>/',
+        'geodata/fmv/<int:pk>/',
         views.FMVEntryDetailView.as_view(),
         name='fmv-entry-detail',
     ),
     path(
-        'geodata/geometries/<int:pk>/',
+        'geodata/geometry/<int:pk>/',
```
banesullivan, line 29: This had been bugging me for a while. It's a small change with little consequence so I included it here rather than making a separate PR. If there is debate on the URLs in general, we should still merge this and make a new issue/PR to clean up these URLs

banesullivan:
@brianhelba, would you please review my changes to the docker-compose setup and how I split the Django and celery images?

rgd/geodata/management/commands/_data_helper.py
```diff
@@ -106,7 +106,11 @@ def load_shape_files(shape_files):
 
 
 def load_fmv_files(fmv_files):
-    raise NotImplementedError('FMV ETL with Docker is still broken.')
+    ids = []
+    for videof in fmv_files:
+        fmv_file = _get_or_create_file_model(models.FMVFile, videof)
+        ids.append(fmv_file.id)
+    return ids
```
zachmullen, line 166:
```suggestion
    return [_get_or_create_file_model(models.FMVFile, fmv).id for fmv in fmv_files];
```

rgd/geodata/models/fmv/etl.py
```diff
@@ -17,59 +16,44 @@
...
-                raise exc
-            logger.info('Finished running image with result %r' % result)
-            # Store result
-            fmv_file_entry.klv_file.save(os.path.basename(output_path), open(output_path, 'rb'))
-            fmv_file_entry.save(
-                update_fields=[
-                    'klv_file',
-                ]
+    with video_file.yield_local_path() as dataset_path:
+        logger.info('Running dump-klv with data %s' % (dataset_path))
```
zachmullen, line 27: FYI, being based on the Girder 4 stack means we can use f-strings:

```python
f'Running dump-klv with data {dataset_path}'
```

I think this is generally preferred as the modern way to do string formatting.

rgd/geodata/models/fmv/etl.py
```diff
@@ -17,59 +16,44 @@
...
+        logger.info('Finished running dump-klv with result %r' % result)
+        # Store result
+        fmv_file_entry.klv_file.save(os.path.basename(output_path), open(output_path, 'rb'))
+        fmv_file_entry.save(
+            update_fields=[
+                'klv_file',
+            ]
+        )
+        shutil.rmtree(tmpdir)
     return
```
zachmullen, line 57: This `return` is also superfluous

rgd/geodata/models/fmv/etl.py
```diff
@@ -17,59 +16,44 @@
...
-        logger.exception('Internal error running dump-klv')
-        raise exc
+            result = 0
+        except subprocess.CalledProcessError as exc:
+            result = exc.returncode
+            logger.info('Failed to successfully run dump-klv (%r)' % (exc))
+            raise exc
+        logger.info('Finished running dump-klv with result %r' % result)
+        # Store result
+        fmv_file_entry.klv_file.save(os.path.basename(output_path), open(output_path, 'rb'))
```
zachmullen, line 50: It looks like this file descriptor isn't getting closed. Let's use a context manager.

rgd/geodata/models/fmv/etl.py
```diff
@@ -17,59 +16,44 @@
...
-            logger.info('Finished running image with result %r' % result)
-            # Store result
-            fmv_file_entry.klv_file.save(os.path.basename(output_path), open(output_path, 'rb'))
-            fmv_file_entry.save(
-                update_fields=[
-                    'klv_file',
-                ]
+    with video_file.yield_local_path() as dataset_path:
+        logger.info('Running dump-klv with data %s' % (dataset_path))
+        tmpdir = tempfile.mkdtemp()
```
zachmullen, line 28: Let's use the [context manager form of this, TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). That will obviate the need to call `shutil.rmtree` later.

rgd/geodata/models/fmv/etl.py
```diff
@@ -243,6 +227,5 @@ def read_fmv_file(fmv_file_id):
     )
 
     _populate_fmv_entry(entry)
-    entry.save()
 
     return
```
zachmullen, line 231: Remove this too

rgd/geodata/models/fmv/etl.py
```diff
@@ -203,9 +187,8 @@ def _convert_video_to_mp4(fmv_file_entry):
 
 
 def _populate_fmv_entry(entry):
-    # Open in text mode
-    with entry.fmv_file.klv_file.open() as file_obj:
-        content = file_obj.read().decode('utf-8')
+    with field_file_to_local_path(entry.fmv_file.klv_file) as path:
+        content = open(path, 'r', encoding='utf-8', errors='ignore').read()
```
zachmullen, line 191: This file descriptor is also unclosed. Use `with open(path, ...)`

rgd/geodata/tests/test_fmv.py
```diff
@@ -1,21 +1,37 @@
...
+
 import pytest
 
 from rgd.geodata.datastore import datastore
 from rgd.geodata.models.fmv.base import FMVEntry
+from rgd.geodata.models.mixins import Status
 
 from . import factories
 
+NO_KWIVER = not shutil.which('kwiver') or not shutil.which('ffmpeg')
```
zachmullen, line 11: If we want to keep these tests more isolated, we could mock the subprocess call. It's up to you though; maybe it's easy enough to install these things in CI.

banesullivan:
When running tests via docker-compose, they must be run through the `celery` image, not the `django` image. This is because the `django` image has just enough deps to run the app but not run the ETL tasks which are tested.

Should adjust the tests that require the `worker` deps list to skip if those are not present? Or should we just leave as is?

zachmullen (changes requested):

rgd/geodata/tests/test_fmv.py
banesullivan: It is easy enough to install these in CI but maybe not on the developer's machine. I figured if someone doesn't have this installed, its easier to skip locally and run in CI

rgd/geodata/models/fmv/etl.py
banesullivan: We do this in a number of places... I will go through and clean up all of these

rgd/geodata/models/fmv/etl.py
banesullivan: This is a bad habit/formatting preference of mine. I will remove

rgd/geodata/models/fmv/etl.py
banesullivan: Leftover from the early days of dev. I will convert all formatted strings to f-strings.

rgd/geodata/tests/test_fmv.py
zachmullen: How about we change this to allow an environment variable to skip these tests, for those who wish to run natively on other platforms? The advantage of that is that we won't ever accidentally skip them in CI if they're explicitly opt-out.

rgd/geodata/management/commands/_data_helper.py
```diff
@@ -159,11 +159,7 @@ def load_shape_files(shape_files):
 
 
 def load_fmv_files(fmv_files):
-    ids = []
-    for videof in fmv_files:
-        fmv_file = _get_or_create_file_model(models.FMVFile, videof)
-        ids.append(fmv_file.id)
-    return ids
+    return [_get_or_create_file_model(models.FMVFile, fmv).id for fmv in fmv_files];
```
zachmullen, line 162: Sorry, not sure why I put a semicolon in my suggestion 😂

rgd/geodata/models/fmv/etl.py
```diff
@@ -17,60 +16,43 @@
...
+        stderr_path = os.path.join(tmpdir, 'stderr.dat')
+        cmd = [
+            'kwiver',
+            'dump-klv',
+            dataset_path,
+        ]
+        try:
+            with open(dataset_path, 'rb') as stdin, open(output_path, 'wb') as stdout, open(
+                stderr_path, 'wb'
+            ) as stderr:
```
banesullivan, line 38: Black did this awful formatting

zachmullen (approved):

PR #313, opened 2021-03-17
Add manage.py command for ingest from S3 bucket

### LINKED ISSUE #302: Point to an S3 bucket and auto-ingest all data
Can we figure how to auto ingest data that match criteria from an existing bucket?

e.g. ingest all GeoTiffs given the bucket location

zachmullen:
Fixes #302


rgd/geodata/management/commands/ingest_s3.py
```diff
@@ -0,0 +1,53 @@
...
+        boto3_params = {
+            'config': botocore.client.Config(
+                signature_version=botocore.UNSIGNED, region_name=region
+            )
+        }
+
+    s3_client = boto3.client('s3', **boto3_params)
+    for obj in _iter_matching_objects(s3_client, bucket, prefix, include_regex):
+        url = f's3://{bucket}/{obj["Key"]}'
+        print(url)  # TODO create database record(s)
```
banesullivan, line 53: This is awesome! I can handle the rest from here using the adjacent `_data_helper` module

zachmullen:
This now supports Google Cloud Storage. Here's an example invocation:

```
./manage.py ingest_s3 gcp-public-data-landsat --prefix LC08/01/044 --google
```

aashish24:
How can I test this feature locally on my RGD instance? Do I need anything special before I run the manage.py?

zachmullen:
If you have it installed via docker-compose, you'd just checkout this branch and run 

```
docker-compose run --rm django ./manage.py ingest_s3 gcp-public-data-landsat --prefix LC08/01/044 --google
```

Then you'd see the listing of the objects in the bucket. Note that it doesn't actually ingest them currently, just iterates over them.

banesullivan:
This is ready to go. It creates a `ChecksumFile` model for every resource found on S3. At this time, this is sufficient for automatically ingesting data.

Down the road, we may want to adjust this such that for every directory found, we create a raster entry of those image files and ancillary data files. 

Unfortunately, even between Landsat and Sentinel, this directory structure varies so we do not want to enforce any sort of structure when ingesting automatically like this as it would be really easy to have that mess up.

For now, this will require a user to manually create ImageFiles and RasterEntrys after ingesting the files.

rgd/geodata/management/commands/ingest_s3.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def _iter_matching_objects(
+    s3_client, bucket: str, prefix: str, include_regex: str
+) -> Generator[dict, None, None]:
+    paginator = s3_client.get_paginator('list_objects')
+    page_iter = paginator.paginate(Bucket=bucket, Prefix=prefix)
+    include_pattern = re.compile(include_regex)
+    # To avoid `$folder$`, etc. objects
+    exclude_pattern = re.compile(r'\$.+\$')
```
zachmullen, line 19: This key name pattern is idiosyncratic to the GCS landsat dataset -- is this script designed to be a general purpose S3 ingest script, or specific to that landsat data? If it's meant to be general, let's make the exclude regex another parameter that can be passed to the command.

rgd/geodata/management/commands/ingest_s3.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _format_url(self, base_url):
+        if self.google:
+            return 'http://storage.googleapis.com/' + base_url
+        return f'https://{self.region}.amazonaws.com/' + base_url
+
+    def load_object(self, obj: dict) -> None:
+        key = obj['Key']
+        url = self._format_url(f'{self.bucket}/{key}')
+        _ = helper._get_or_create_checksum_file_url(url, name=obj['Key'])
```
zachmullen, line 41:
```suggestion
        helper._get_or_create_checksum_file_url(url, name=obj['Key'])
```

rgd/geodata/management/commands/ingest_s3.py
```diff
@@ -0,0 +1,81 @@
...
+class CloudLoader:
+    def __init__(self, bucket: str, region: str, google: bool = False):
+        self.bucket = bucket
+        self.google = google
+        self.region = region
+
+    def _format_url(self, base_url):
+        if self.google:
+            return 'http://storage.googleapis.com/' + base_url
+        return f'https://{self.region}.amazonaws.com/' + base_url
```
zachmullen, line 36: Prefer f-strings to `+` concatenation

rgd/geodata/management/commands/ingest_s3.py
banesullivan: This is intended to be a general-purpose S3 ingest script, so fair enough... I'll compromise and set `\$.+\$` as the default exclude pattern for now. Folks will forget to include it otherwise and it'll result in errors

zachmullen: That seems more like a matter for documentation; having a default behavior of excluding any key with at least two non-consecutive `$` characters in it seems to violate the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) IMO. If we really want an easier-to-use special case for the landsat dataset specifically, I believe that should be a separate command whose semantics correspond to the domain model of the landsat dataset, rather than that of S3. Curious what others think, maybe @brianhelba

banesullivan: I could change the regex to `\$\w+\$` such that it isn't any two non-consecutive `$` but any word surrounded by `$` to reduce the astonishment factor

banesullivan: Or we could leave the exclude regex totally optionally and hardcode to avoid `$folder$` with `.endswith('$folder$')`. I cannot find any documentation for these folder objects on the GCS landsat data

brianhelba: I see that `--include-regex` is already a CLI option. Could the same be done for `--include-regex`? This ought to facilitate the capability while avoiding hardcoding anything that's idiosyncratic to one data set.

brianhelba: I see the concern about end users forgetting to include it. I think documentation, particularly working examples, would be a good solution. The category of user which we're trying to defensively engineer around is likely to copy-paste any commands from the docs, rather than building up all the arguments thoughtfully.

As long as you give those users the correct examples (including `--exclude-regex`) to copy-paste, I suspect they'll be able to do so without altering anything.

banesullivan: Fair enough. I will add some documentation of examples and leave `--exclude-regex` without any default option

PR #315, opened 2021-03-18
Fix index URL redirect

banesullivan:
We had the base URL redirecting to `geodata/spatial_entries/`. This removes that and puts the spatial entries list view/search page as the homepage.


zachmullen (approved):
If we care about preserving URLs, we could keep the old URL and redirect it to `/`, but I am guessing we don't care about that at this point.

banesullivan:
> but I am guessing we don't care about that at this point.

We sure don't!

PR #324, opened 2021-03-19
Allow copying GeoJSON to clipboard

### LINKED ISSUE #321: Copy GeoJSON button not working
The "Copy GeoJSON" button on the detail page is not currently working (for me in Chrome)

The button is here: 

https://github.com/ResonantGeoData/ResonantGeoData/blob/928a06e810faffa5a221f50e6430082c13c249a7/rgd/geodata/templates/geodata/_base/detail.html#L16-L30

And the text area it is supposed to copy is here:

https://github.com/ResonantGeoData/ResonantGeoData/blob/928a06e810faffa5a221f50e6430082c13c249a7/rgd/geodata/templates/geodata/_base/detail.html#L38-L40

zachmullen:
Fixes #321

~~Sorry for the aggressive reformatting that my IDE did. Is there some VS Code plugin for styling these pages that you recommend? I am not sure if this file was auto-formatted before.~~

EDIT: Answered my own question, the Django extension handles this nicely. I fixed the formatting issues.

### DIFF
--- a/rgd/geodata/templates/geodata/_base/detail.html
+++ b/rgd/geodata/templates/geodata/_base/detail.html
@@ -17,7 +17,6 @@
     <span class="text-has-dark-bg is-block">Acquired:</span>
     {{ object.acquisition_date }}
   </div>
-  {% comment %}
   <div class="column is-2">
     <a id="copyjson" class="button has-icon is-flex is-justify-content-space-between is-secondary is-fullwidth">
       <span>Copy GeoJSON</span>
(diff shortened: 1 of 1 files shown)


zachmullen:
This provides a quick fix for the issue, however long term I'd prefer this to go away. I think as a rule, anytime we have a "copy to clipboard" button, it should be next to a readonly input element that contains the text itself, to provide a manual backup option in any cases when the `execCommand` doesn't work or is disabled.

banesullivan (approved):

PR #326, opened 2021-03-23
Add seperate worker instance and task queue

banesullivan:
cc @zachmullen for #285 

Here are some experimental changes I had for adding a separate worker instance and task queue but it isn't quite working

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -27,7 +27,7 @@ services:
       "worker",
       "--loglevel", "INFO",
       "--without-heartbeat",
-      "-Q default"
+      "-X priority"
     ]
     # Docker Compose does not set the TTY width, which causes Celery errors
     tty: false
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -26,7 +26,34 @@ services:
       "--app", "rgd.celery",
       "worker",
       "--loglevel", "INFO",
-      "--without-heartbeat"
+      "--without-heartbeat",
+      "-Q default"
+    ]
+    # Docker Compose does not set the TTY width, which causes Celery errors
+    tty: false
+    privileged: true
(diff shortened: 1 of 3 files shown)


zachmullen:
What's the problem? (FYI @brianhelba )

banesullivan:
The problem is tasks are launching or being received... I imagine this is a trivial fix for something I am forgetting to set

rgd/celery.py
```diff
@@ -11,6 +11,7 @@
 # Using a string config_source means the worker doesn't have to serialize
 # the configuration object to child processes.
 app = Celery('rgd', config_source='django.conf:settings', namespace='CELERY')
+app.conf.task_default_queue = 'default'
```
brianhelba, line 14: All the other Celery config is done via Django settings. I think you should set that there, for consistency.

docker-compose.override.yml
```diff
@@ -26,7 +26,34 @@ services:
       "--app", "rgd.celery",
       "worker",
       "--loglevel", "INFO",
-      "--without-heartbeat"
+      "--without-heartbeat",
+      "-Q default"
```
brianhelba, line 30: You could use `-X` / `--exclude-queues` instead: https://docs.celeryproject.org/en/stable/reference/cli.html#cmdoption-celery-worker-X

rgd/celery.py
banesullivan: With `--exclude-queues` I don't think this queue is needed.

banesullivan:
@brianhelba @zachmullen, I cannot get the tasks with `queue='priority'` to be received by the `celery_priority` container. All other tasks are launching in the default `celery` container though

banesullivan:
Closing due to inactivity and other priorites

PR #332, opened 2021-03-26
Add Python Client

### LINKED ISSUE #178: Python API from OpenAPI endpoints
We need to (automatically?) generate a Python API from the OpenAPI endpoints. Or at least provide a way to access those endpoints from a Python client.

One viable option is: https://github.com/OpenAPITools/openapi-generator

Another approach would be to manually create the API... I'm actually leaning towards this to make sure we can design the API in a way that is user friendly and have clear documentation

First, we need a few clear use cases, likely coming from ndsampler

### LINKED ISSUE #305: Stub out Client Library
Initial task for #178

AlmightyYakob:
Fixes #305 resolve #178

This adds an `rgdc` package, which contains an `Rgdc` class. This class is the client class, and contains all the relevant methods for interacting with RGD.

This is just some initial methods, based on #178. Hopefully what I have addresses the use cases correctly, although I'll admit I'm a bit distanced from them.

This PR does not include tests (although I would really like to add them), as that would involve spinning up a test RGD instance in CI. This presents issues regarding both feasibility (e.g. system packages), as well as performance. For now, I've decided to punt on it.


aashish24:
@AlmightyYakob this is great. I see an example in the readme. It would be desired to have an example that first get a list of ids based on some filter (cloud mask, aoi) and then download these tiles. Is it possible to update your example to have these pieces?

AlmightyYakob:
> @AlmightyYakob this is great. I see an example in the readme. It would be desired to have an example that first get a list of ids based on some filter (cloud mask, aoi) and then download these tiles. Is it possible to update your example to have these pieces?

What endpoint would I use to get the list of IDs? The `search` endpoint only returns SpatialEntry IDs.

aashish24:
SpatialEntry IDs are fine.

banesullivan:
This is waiting on me.

I need to add a few new endpoints to better serialize raster entries, search raster entries specifically, and a few other things

Currently there is no good route to go from the IDs returned by the search endpoint to downloading image files

aashish24:
In the future we want a) given AOI, b) some facet parameter, and c) data type information (raster or vector), the system generates  a list of raster entry ids (in case data type from input is raster type) or vector entry ids, then she can filter based on some criterions by looping over these ids, and finally ask a subset of ids to be downloaded to her desktop with all other metadata files.

rgdc/rgdc/types.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
+from typing import Literal, Tuple, Union
```
banesullivan, line 2: `Literal` is introduced in Python 3.8: https://www.python.org/dev/peps/pep-0586/

rgdc/setup.py
```diff
@@ -0,0 +1,43 @@
...
+    author_email='kitware@kitware.com',
+    keywords='',
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python',
+    ],
+    python_requires='>=3.7',
```
banesullivan, line 35: Python 3.8 minimum because of `Literal` type

https://github.com/ResonantGeoData/ResonantGeoData/pull/332/files?short_path=e73c802#r609136159

rgdc/rgdc/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.0'
```
banesullivan, line 1: Moving forward, should we have this version match RGD's version?

rgdc/rgdc/rgdc.py
```diff
@@ -0,0 +1,248 @@
...
+        if query:
+            if isinstance(query, str):
+                try:
+                    query = json.loads(query)
+                except JSONDecodeError:
+                    pass
+
+            if isinstance(query, dict):
+                # Allow failure on invalid format
+                query = wkt.dumps()
```
banesullivan, line 201: is this supposed to be?:

```suggestion
                query = wkt.dumps(query)
```

rgdc/rgdc/rgdc.py
```diff
@@ -0,0 +1,248 @@
...
+            file_path = parent_path / filename
+            with open(file_path, 'wb') as open_file_path:
+                for chunk in iterate_response_bytes(file_download_url):
+                    open_file_path.write(chunk)
+
+        return str(path)
+
+    def search(
+        self,
+        query: Optional[Union[Dict, str]] = None,
```
banesullivan, line 137: The query geometry is optional but `None` is an invalid default type for it:

```
ValueError: Cannot use None as a query value
```

this is a server error and we get a 500 status back but the client does not raise an exception. this is a problem.

rgdc/rgdc/utils.py
```diff
@@ -0,0 +1,47 @@
...
+            url = r.links['next']['url']
+        else:
+            break
+
+
+def results(responses: Iterator[Response]) -> Iterator[dict]:
+    for response in responses:
+        response.raise_for_status()
+        for result in response.json()['results']:
+            yield result
```
banesullivan, line 26: This is pretty cool. However, I don't think it makes sense to use a generator on the user side. If doing so, you have to rerun the request if you want to access a result more than once.

Or you have to store the results of the generator yourself:

```py
q = client.search(query=json.dumps(bbox), predicate='within')
persistent = [e for e in q]
```

This is a bit cumbersome when I typically might want to iterate over the query more than once

```py
for result in q:
    print(q)
    ... # do something
    
for result in q: # <-- q is empty at this point
    print(q)
    ... # do something else
```

rgdc/rgdc/rgdc.py
```diff
@@ -0,0 +1,248 @@
...
+            parent_path = path / parent_dirname if parent_dirname else path
+            parent_path.mkdir(parents=True, exist_ok=True)
+
+            # Download contents to file
+            file_path = parent_path / filename
+            with open(file_path, 'wb') as open_file_path:
+                for chunk in iterate_response_bytes(file_download_url):
+                    open_file_path.write(chunk)
+
+        return str(path)
```
banesullivan, line 133: I'm thinking it may be useful to return a structure with all of the paths to the files downloaded by this. Perhaps a dict with the keys `images` and `ancillary` which are lists of the paths to each of those files?

rgdc/rgdc/version.py
AlmightyYakob: I'd say so yes, but until we have a good feature match, we'll want to keep the client versions to prereleases, so that adding additional features don't misalign the versions.

banesullivan:
@AlmightyYakob, I'm wrapped up with this. Would you do a final look over at my changes and let me know when you think this is good to merge?

banesullivan (approved):

banesullivan (approved):

PR #333, opened 2021-03-26
Google Cloud Landsat & Sentinel Data Command

### LINKED ISSUE #301: Ingest existing indexes for Landsat data
Add a way to pass an existing data catalog index from Google cloud to populate Landsat data

*more details to come*

### LINKED ISSUE #317: Auto ingest from existing Google Cloud catalogs

banesullivan:
Provides various improvements to the demo data commands and adds a new one for importing Landsat and Sentinel imagery from Google Cloud

Note that I am not pointing to the full GC index files but a subset of it that I have added to DKC:
- Landsat: https://data.kitware.com/#item/605cbf9c2fa25629b9e15c5c
- Sentinel: https://data.kitware.com/#item/6064f7e92fa25629b9319906

This subset index file is in the same format as the original index on GC but only has imagery for 2015 through 2019 over the Korea site (note that it is only `MGRS_TILE` `52SDG` for the sentinel data for now)

In a comment below, I will add instructions for creating a custom subset using GeoJSON geometry

cc @aashish24 @matt-bernstein

close #301 close #317


rgd/geodata/management/commands/gc_landsat.py
```diff
@@ -0,0 +1,145 @@
...
+            'B6_VCID_1.TIF',
+            'B6_VCID_2.TIF',
+            'B7.TIF',
+            'B8.TIF',
+            'B9.TIF',
+            'BQA.TIF',
+        ]
+        possible_anc = [
+            'MTL.txt',
+        ]
```
banesullivan, line 98: Needs double checking

banesullivan: @matt-bernstein, could you do a sanity check on this?

matt-bernstein: @banesullivan
This ingestion shouldn't touch `sensor == 'TM'`, as that's from Landsat 4-5, which were discontinued in 2013. You probably also want to move the non-`TIF` files from `possible_bands` to `possible_anc`.

For Landsat 7 and 8, there is another ancillary file `ANG.txt`. (ex. [L7](https://console.cloud.google.com/storage/browser/gcp-public-data-landsat/LE07/01/001/001/LE07_L1GT_001001_20010715_20170204_01_T2), [L8](https://console.cloud.google.com/storage/browser/gcp-public-data-landsat/LC08/01/001/002/LC08_L1GT_001002_20160817_20170322_01_T2)) I am not sure why this file was missing in the scenes I downloaded using [FeLS](https://gitlab.kitware.com/smart/fetchLandsatSentinelFromGoogleCloud) and [Landsat578](https://gitlab.kitware.com/smart/Landsat578). Maybe it only shows up sometimes. It does look like a valid angle map to me.

For Landsat 8, this is otherwise correct for `sensor == 'OLI_TIRS'`

The sensors `OLI` and `TIRS` alone ([ex](https://console.cloud.google.com/storage/browser/gcp-public-data-landsat/LO08/01/002/010/LO08_L1GT_002010_20150211_20170413_01_T2), [ex](https://console.cloud.google.com/storage/browser/gcp-public-data-landsat/LT08/01/001/248/LT08_L1GT_001248_20130318_20170505_01_T2)) contain only B1-9 and B10-11 respectively. These are the visual vs. thermal bands. 

These 3 sensor types for Landsat 8 map to `LC08/`, `LO08/`, and `LT08/` respectively - see [here](https://gitlab.kitware.com/smart/Landsat578#-1) for nomenclature.

For now I would not download `OLI` or `TIRS` alone. There is probably a good reason the universities and FeLS/Landsat578 ignore them, though I will follow up next week to confirm.

banesullivan: Thanks for the review!

banesullivan:
To create a custom subset of the index, use the attached Jupyter notebook with the original Google Cloud index CSV file with the shapefile of the Landsat tiles

[GC_Landsat_Curate.ipynb.zip](https://github.com/ResonantGeoData/ResonantGeoData/files/6214438/GC_Landsat_Curate.ipynb.zip)


(I will keep this better updated locally)

PR #334, opened 2021-04-01
More robust dev server configuration for demoing

zachmullen:
This uses gunicorn with 8 concurrent threads to make for a better demo from someone's local setup. This makes tile serving somewhat faster, and also partly mitigates the page refusing to reload issue. It also disables debug mode on the server side, which could have performance benefits, but I have not actually measured those. Prior to this change, the server could only handle a single request at a time in dev mode. Using the threaded mode of gunicorn should also prevent subverting our caching behavior.

### DIFF
--- a/dev/.env.docker-compose
+++ b/dev/.env.docker-compose
@@ -1,4 +1,4 @@
-DJANGO_CONFIGURATION=DevelopmentConfiguration
+DJANGO_CONFIGURATION=NonDebugDevConfiguration
 DJANGO_DATABASE_URL=postgres://postgres:postgres@postgres:5432/django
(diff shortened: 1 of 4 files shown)


zachmullen:
I've made this a draft as it's not intended to be merged in the current state. Until we figure out a better plan, this commit can be cherry-picked onto some other branch when someone is preparing a demo.

banesullivan:
Redone in https://github.com/ResonantGeoData/RD-OpenGeo/pull/12

zachmullen:
This has a typo, it should be `DEBUG = False`. I've force pushed the `local-demo-config` branch to contain that change, so use that rather than this PR.

PR #335, opened 2021-04-02
API Improvements

banesullivan:
For #332 

Adds new serializers/endpoints to make it easier to perform a search and go straight to downloading data

This splits up the `geodata` endpoints into three new tags:

- `geodata`: for accessing models via serializer and downloading raw data
- `geosearch`: for performing searches against data in the catalog
- `geoprocess`: for performing operations on data in the catalog (e.g. subsampling, COG conversion, tiling)

I updated the demo notebook (`demos/GeoData-API.ipynb`) to show what is now possible - this is a developer's demo and not intended to be shown externally. @AlmightyYakob, this should be everything needed to proceed with #332

### DIFF
--- a/demos/GeoData-API.ipynb
+++ b/demos/GeoData-API.ipynb
@@ -14,7 +14,8 @@
     "from rasterio.plot import show\n",
     "import matplotlib.pyplot as plt\n",
     "import numpy as np\n",
-    "import json"
+    "import json\n",
+    "from requests.auth import HTTPBasicAuth"
    ]
   },
   {
@@ -25,8 +26,8 @@
    "source": [
     "class GeoData:\n",
     "    HOST = 'http://localhost:8000'\n",
-    "    def __init__(self,):\n",
-    "        pass\n",
+    "    def __init__(self, user, password):\n",
+    "        self.auth=HTTPBasicAuth(user, password)\n",
     "\n",
     "    @staticmethod\n",
     "    def _load_image(content):\n",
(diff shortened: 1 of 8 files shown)


AlmightyYakob:
I appear to get an off-by-one error when comparing the `subentry_pk` and `detail` fields of a geosearch result. Example:

```
      ...
      "subentry_type": "RasterMetaEntry",
      "subentry_pk": 7,
      "subentry_name": "20091021202517-01000100-VIS_0001.ntf",
      "detail": "http://localhost:8000/api/geodata/imagery/raster_entry/6"
```

Notice how `subentry_pk` is `7`, while the raster entry referenced in `detail` has a pk of `6`.

banesullivan:
>Is this due to the difference between RasterEntry and RasterMetaEntry?

Yes, the spatial entry type is RasterMetaEntry and the `"detail"` link points to RasterEntry.

I will fix this

banesullivan:
Good catch

banesullivan:
Now the `SpatialEntry` serializer points directly to `RasterMetaEntry` and that serializer holds the `RasterEntry` to avoid confusion on the client side. This is consistent with our approach for `ImgeFile`/`ImageEntry`, `GeometryArchive`/`GeometryEntry`, and `FMVFile`/`FMVEntry`.

The following are subtypes of `SpatialEntry` and accessible via the API: `FMVEntry`, `GeometryEntry`, and `RasterMetaEntry`

The following are prerequisite, user-generated models for the above (respectfully): `FMVFile`, `GeometryArchive` and `RasterEntry`

Each of the "prerequisite" models are serialized under the spatial entry subtypes and not directly accessible via the API. At a later time, we can expose POST endpoints for these models nut not GET.

The naming choice for `RasterMetaEntry`/`RasterEntry` is a bit confusing/inconsistent with the other models but the logic is the same.

PR #345, opened 2021-04-09
Upgrade to cookiecutter 0.18

brianhelba:


aashish24 (commented):
Cool, does it mean that now we will be using github actions?

brianhelba:
> does it mean that now we will be using github actions?
Yes, this will move RGD to use GitHub actions instead of CircleCI. I find GitHub actions to have a much better integrated interface, and it will be easier for extending to automate other actions in the future, like auto-release of a client package.

@ResonantGeoData/developers However, when this is merged, all existing PRs will need to be rebased onto `master` or have `master` merged into them.

brianhelba:
@banesullivan This is ready for review. If it seems ok, when would be a less disruptive time to merge?

setup.py
```diff
@@ -58,7 +57,7 @@
         'rules',
         'uritemplate',
         # Production-only
-        'django-composed-configuration[prod]',
+        'django-composed-configuration[prod]>=0.16',
```
banesullivan, line 60: See L67 too where `django-composed-configuration[dev]` is listed. I don't think setup tools will grab the version from the default section in `dev` mode

banesullivan:
I'm good merging this whenever. Rebasing open PRs shouldn't be too cumbersome

banesullivan (approved):

PR #347, opened 2021-04-12
Add url to setup.py

Erotemic:
Pypi doesn't have a link to the github webpage. This fixes that.


banesullivan (approved):

banesullivan:
This is actually the wrong setup.py file, but I will update the other to have this

(rgdc is a separate package from rgd)

PR #350, opened 2021-04-12
Add AWS CNAME record for WATCH

banesullivan:


brianhelba (approved):
Maybe a better file name is `watch.tf`. Everything in Terraform is for some sort of deployment.

Otherwise, LGTM.

Issue #354, opened 2021-04-13
Search endpoint returning duplicate SpatialEntries


banesullivan:

matt-bernstein:
A detailed replication is below. cc @banesullivan @AlmightyYakob 

There are 3 separate issues going on here:
- behavior of "limit" kw
- "acquired" kw not functioning
- each file in a RasterMetaEntry as a separate search result

I am not sure which can be fixed in `rgdc` versus RGD itself. I would be happy to get to work on this once we start figuring that out.

```python
import os
import json
import numpy as np
from datetime import date, datetime, timedelta

from fels import run_fels, safedir_to_datetime, landsatdir_to_date

from rgdc import Rgdc

# pick an AOI

top, left = (128.6643, 37.6601)
bottom, right = (128.6749, 37.6639)

# https://landsat.usgs.gov/ard_tile
LS_path_row = (115, 34)  # the Landsat tile that contains the AOI
# https://mappingsupport.com/p2/coordinates-mgrs-gissurfer-maps.html
S2_tile = '52SDG'  # the Sentinel-2 tile that contains the AOI

# and a date range of 1 week

dt_min, dt_max = (datetime(2018, 11, 1), datetime(2018, 11, 8))

# compare to FetchLandsatSentinel
# https://gitlab.kitware.com/smart/fetchLandsatSentinelFromGoogleCloud


def search_tiles_fels():
    '''
output:

Sentinel-2:
Searching for Sentinel-2 images in catalog...
Found 3 files.
['http://storage.googleapis.com/gcp-public-data-sentinel-2/tiles/52/S/DG/S2B_MSIL1C_20181106T020849_N0207_R103_T52SDG_20181106T034331.SAFE', 'http://storage.googleapis.com/gcp-public-data-sentinel-2/tiles/52/S/DG/S2A_MSIL1C_20181101T020821_N0206_R103_T52SDG_20181101T040328.SAFE', 'http://storage.googleapis.com/gcp-public-data-sentinel-2/tiles/52/S/DG/S2A_MSIL1C_20181104T021841_N0206_R003_T52SDG_20181104T055443.SAFE']
[datetime.datetime(2018, 11, 6, 2, 8, 49), datetime.datetime(2018, 11, 1, 2, 8, 21), datetime.datetime(2018, 11, 4, 2, 18, 41)]
Landsat-7:
Searching for Landsat-ETM images in catalog...
Found 1 files.
[...]

PR #355, opened 2021-04-14
Move instrumentation to be SpatialEntry field

banesullivan:
Resolves part of #354 

Now all `SpatialEntry`s can be search against `instrumentation`


banesullivan:
@matt-bernstein, this is the last bit to resolve #354

banesullivan:
Merging due to deliverable timelines

PR #356, opened 2021-04-14
Remove rastermetaentry created check in acquisition_date filter

banesullivan:
Resolves part of #354

### DIFF
--- a/rgd/geodata/filters.py
+++ b/rgd/geodata/filters.py
@@ -61,7 +61,6 @@ class SpatialEntryFilter(filters.FilterSet):
         field_name='acquisition_date',
         help_text='The ISO 8601 formatted date and time when data was acquired.',
         label='Acquired',
-        method='filter_acquired',
     )
     created = filters.IsoDateTimeFromToRangeFilter(
         field_name='created',
@@ -163,29 +162,6 @@ class SpatialEntryFilter(filters.FilterSet):
             return queryset.filter(rastermetaentry__isnull=False)
         if value == 'fmv':
             return queryset.filter(fmventry__isnull=False)
-        return queryset
-
-    def filter_acquired(self, queryset, name, value):
-        """Filter by when the data was acquired.
-
-        There is a special where filtering by acquisition, created is also filtered
-        for `RasterEntry`.
-        """
-        if value:
-            if value.start is not None:
-                queryset = queryset.filter(
-                    (
-                        Q(acquisition_date__gte=str(value.start))
-                        | Q(rastermetaentry__parent_raster__created__gte=str(value.start))
-                    )
-                )
-            if value.stop is not None:
-                queryset = queryset.filter(
-                    (
-                        Q(acquisition_date__lte=str(value.stop))
-                        | Q(rastermetaentry__parent_raster__created__lte=str(value.stop))
-                    )
-                )
         return queryset
 
     def filter_created(self, queryset, name, value):


mcovalt:
I was a bit fuzzy on this from the get-go. There's a discussion in this review: https://github.com/ResonantGeoData/ResonantGeoData/pull/303#discussion_r591962106. I'm interested in seeing this chunk of code go as a whole, however, I do want to understand the original logic before we nix it.

In [`rgd.geodata.api.search.extent_summary`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L418-L437), the flag `has_created` makes it so that the value of `acquisition_date` is [possibly sourced from `created`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L424-L425). This flag is only ever passed when searching rasters (see [1](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L499), [2](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L541), [3](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L499)).

The `filter_acquired` method being dropped here attempts to replicate that logic. Am I understanding what was originally happening in [`rgd.geodata.api.search.extent_summary`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L418-L437)?

banesullivan:
> I was a bit fuzzy on this from the get-go. There's a discussion in this review: [#303 (comment)](https://github.com/ResonantGeoData/ResonantGeoData/pull/303#discussion_r591962106). I'm interested in seeing this chunk of code go as a whole, however, I do want to understand the original logic before we nix it.
> 
> In [`rgd.geodata.api.search.extent_summary`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L418-L437), the flag `has_created` makes it so that the value of `acquisition_date` is [possibly sourced from `created`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L424-L425). This flag is only ever passed when searching rasters (see [1](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L499), [2](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L541), [3](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L499)).
> 
> The `filter_acquired` method being dropped here attempts to replicate that logic. Am I understanding what was originally happening in [`rgd.geodata.api.search.extent_summary`](https://github.com/ResonantGeoData/ResonantGeoData/blob/44e73da9851e80d7326c32e26ac126b6164be613/rgd/geodata/api/search.py#L418-L437)?

@mcovalt, that's right. I believe we initially did this because we had no way of auto-populating the acquisition dates and they were almost always null. At the time, I think we implemented it this way so that the acquisition date would fall back to when the record was created since they were null but that is now quite problematic since we are auto-populating many acquisition dates for rasters... or at least the OR logic here is problematic. Either way, this needs to go.

mcovalt (approved):

PR #359, opened 2021-04-14
Remove paginator and limits on search endpoint

banesullivan:
Simple solution to the limit issue in #354 

This removes the paginator and `limit` argument

cc @matt-bernstein

### DIFF
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -183,7 +183,6 @@ class Rgdc:
         resolution: Optional[Tuple[int, int]] = None,
         cloud_cover: Optional[Tuple[float, float]] = None,
         frame_rate: Optional[Tuple[int, int]] = None,
-        limit: Optional[int] = None,
         offset: Optional[int] = None,
     ) -> Iterator[Dict]:
         """
@@ -208,7 +207,6 @@ class Rgdc:
             resolution: The min/max resolution of the raster.
             cloud_cover: The min/max cloud coverage of the raster.
             frame_rate: The min/max frame rate of the video.
-            limit: The maximum number of results to return.
             offset: The number of results to skip.
 
         Returns:
@@ -221,7 +219,6 @@ class Rgdc:
             'relates': relates,
             'datatype': datatype,
             'instrumentation': instrumentation,
-            'limit': limit,
             'offset': offset,
         }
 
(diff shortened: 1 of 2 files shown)


rgdc/rgdc/rgdc.py
```diff
@@ -183,7 +183,6 @@ def search(
         resolution: Optional[Tuple[int, int]] = None,
         cloud_cover: Optional[Tuple[float, float]] = None,
         frame_rate: Optional[Tuple[int, int]] = None,
-        limit: Optional[int] = None,
         offset: Optional[int] = None,
```
zachmullen, line 186: Limit and offset typically go together. If our intent is to disable all pagination capabilities, we should remove both of them.

zachmullen (changes requested):

rgdc/rgdc/rgdc.py
banesullivan: Ah thanks for catching that

banesullivan:
@zachmullen, is this a bad idea to not limit the search results?

I could see this possibly blowing up but we need #354 resolved ASAP for a demo and current use cases do not have a ton of data

zachmullen:
>  is this a bad idea to not limit the search results?

I think removing pagination altogether is the wrong solution for the problem as stated. For results sets that contain thousands of records, using paging is a very appropriate approach. What I would do is have the client library handle paging internally in a way that's transparent to the end user, perhaps with a generator as the API.

matt-bernstein:
After realizing what `offset` was for, I was able to make `rgdc.Rgdc.search` behave correctly with the following snippet:

```python
    def _search(instrumentation):
        offset = 0
        results = []
        while (query := client.search(**kwargs, instrumentation=instrumentation, offset=offset)):
            # fix dates
            query = ([
                entry for entry in query if (dt_min <= datetime.fromisoformat(
                    entry['acquisition_date'].strip('Z')) <= dt_max)
            ])
            # fix dupes
            _, ixs = np.unique([entry['detail'] for entry in query],
                               return_index=True)
            query = list(np.array(query)[ixs])

            results.extend(query)

            offset += kwargs['limit']

        return results
```
and I agree that pagination is useful for these very large result sets (which will cease to be very large when the other issues in #354 are fixed, but other use cases may continue to be).

I'd support something like having `rgdc.Rgdc.search` do this automatically when `limit=None`.

banesullivan:
Closing this in favor of #360 and #362

PR #361, opened 2021-04-15
Add STAC data model

mcovalt:
The main challenge in this part of the project was adapting the relatively unstructured STAC model to a structured schema.

Instead of imposing a rigid structure on the STAC data model, a framework to safely extend it is introduced. The result is a (hopefully) compliant and flexible library.

The top-level STAC models implemented here are very sparse. They capture the necessary structure and relationships of the data model, though. This is intentional. The full data model is added onto it in the `extensions` folder.


mcovalt:
Rather than rebasing our current database schema onto a new database schema optimized for STAC workloads, we plan to implement schema crosswalks between our current database schema and the domain objects defined in the STAC specification. These schema crosswalks will likely be defined in serializers scoped to produce fixed views of our data that are STAC compliant.

PR #360, opened 2021-04-14
handle pagination in Rgdc.search

matt-bernstein:
Solves part of #354 , replaces #359 

```Rgdc.search(limit=None)``` now reads and returns all pages instead of silently defaulting to `limit=100`.

PR #362, opened 2021-04-15
Add pagination to `Rgdc.search`

### LINKED ISSUE #354: Search endpoint returning duplicate SpatialEntries

AlmightyYakob:
Closes #354, Supersedes #359.

This conflicts with #360, although I was unaware of this PR until now. However, I think my PR addresses slightly more use cases than that one.

Currently, this function will fetch as many results as it can from the server at once, and paginate until the user specified limit is reached. So, for example, if the user specifies a limit of 1200, this function will first just pass that limit to the server, which will only return 1000 results, since that's the silent max. This function will then fetch the remaining 200 results after that.

However, I've also considered modifying this function to always use the server default of 100, and deal with the limit solely in the client. The main reason for this would be to keep individual request times down, as for 1000 results, the requests can take around 20 seconds. This would mean a slightly higher overall time for fetching the same amount of data, but would also allow for a more flexible use of the data. Thoughts @banesullivan @matt-bernstein?

*Closes #360*

### DIFF
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -217,7 +217,7 @@ class Rgdc:
             offset: The number of results to skip.
 
         Returns:
-            An list of Spatial Entries.
+            A list of Spatial Entries.
         """
         # The dict that will be used to store params.
         # Initialize with queries that won't be additionally processed.
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -11,7 +11,12 @@ from tqdm import tqdm
 
 from .session import RgdcSession
 from .types import DATETIME_OR_STR_TUPLE, SEARCH_DATATYPE_CHOICE, SEARCH_PREDICATE_CHOICE
-from .utils import DEFAULT_RGD_API, datetime_to_str, download_checksum_file_to_path
+from .utils import (
+    DEFAULT_RGD_API,
+    datetime_to_str,
+    download_checksum_file_to_path,
+    limit_offset_pager,
+)
 
 
 @dataclass
@@ -281,6 +286,4 @@ class Rgdc:
             params['frame_rate_min'] = frmin
             params['frame_rate_max'] = frmax
 
-        response = self.session.get('geosearch', params=params)
-        response.raise_for_status()
-        return [result for result in response.json()['results']]
+        yield from limit_offset_pager(self.session, 'geosearch', params=params)
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -190,7 +190,7 @@ class Rgdc:
         frame_rate: Optional[Tuple[int, int]] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
-    ) -> Iterator[Dict]:
+    ) -> List[Dict]:
         """
         Search for geospatial entries based on various criteria.
 
@@ -286,4 +286,4 @@ class Rgdc:
             params['frame_rate_min'] = frmin
             params['frame_rate_max'] = frmax
 
-        yield from limit_offset_pager(self.session, 'geosearch', params=params)
+        return list(limit_offset_pager(self.session, 'geosearch', params=params))
--- a/rgdc/rgdc/utils.py
+++ b/rgdc/rgdc/utils.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from pathlib import Path
-from typing import Dict, Iterator, List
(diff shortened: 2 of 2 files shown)


matt-bernstein:
My preferred behavior on the client's end is to do away with the silent max on `limit` entirely - the default should be to return all results. The user shouldn't need to care or guess how many results there might be by setting `limit` in all cases, as this is a potential source of errors if they don't check that `limit` has been reached.

I wouldn't worry too much about having large request times, since after the other duplication issues in #354 are fixed, a typical query would be ~10 entries and a soft maximum would be ~1000. Unless there's some future use case which involves querying the entire globe at once in a single process, which sounds like a bad idea anyway.

AlmightyYakob:
> My preferred behavior on the client's end is to do away with the silent max on `limit` entirely - the default should be to return all results. The user shouldn't need to care or guess how many results there might be by setting `limit` in all cases, as this is a potential source of errors if they don't check that `limit` has been reached.

I'm a bit confused what you mean by this. There's no way client side to do away with the silent max. Are you saying the `limit` parameter isn't useful for this function?

matt-bernstein:
On the user's end, if that wasn't clear. As in, if `limit=None`, all results should be returned, even if `search()` has to silently make multiple requests and handle pagination to make that work.

AlmightyYakob:
> On the user's end, if that wasn't clear. As in, if `limit=None`, all results should be returned, even if `search()` has to silently make multiple requests and handle pagination to make that work.

Oh, yeah I agree, that's how my implementation currently operates. My concern was more about the size of each indvidivual request when that's the case, but as you indicated I suppose it's not a real issue.

rgdc/rgdc/rgdc.py
```diff
@@ -281,6 +286,4 @@ def search(
             params['frame_rate_min'] = frmin
             params['frame_rate_max'] = frmax
 
-        response = self.session.get('geosearch', params=params)
-        response.raise_for_status()
-        return [result for result in response.json()['results']]
+        yield from limit_offset_pager(self.session, 'geosearch', params=params)
```
matt-bernstein, line 289: I think this is the change I'm referring to. This would turn `search()` into a generator, right? Something like
```python
result = []
for page in limit_offset_pager(self.session, 'geosearch', params=params):
    result.extend(page)
return result
```

Would preserve the original call signature and be nicer, imo.

matt-bernstein: I think that's the only significant difference from #360

AlmightyYakob: Yes, with my change it's a generator. My reasoning was that if you wanted to use the results in a generator fashion then it's useful, and if not, you can always just cast the return of `search` to a list. If the consensus from you and @banesullivan is that it should return a list, then I'm happy to change it.

matt-bernstein: Yeah, I'd be happy to close #360 with that change

AlmightyYakob: FWIW, another difference between this and #360 is that if a limit above the silent server limit is specified (say 1200), then in the implementation in #360, only the server limit of 1000 is returned, because pagination is only done if limit is `None`.

banesullivan: I agree that returning a generator might not be very user-friendly. It may be useful to have a keyword argument `return_generator=False` that will return the generator when requested but default to casting it to a list.

I want to keep the generator under the hood in case we implement downstream functionality that could benefit from it but for most use cases right now, we'd be best off with a static list of the results.

AlmightyYakob: It looks like you both agree, so I'll change it

AlmightyYakob: Turns out it's impossible for a function to conditionally return a list / act as a generator, so I've just returned a list for now. If the need arises we can expose a new function that operates the same way but produces a generator instead. The generator functionality is all still under the hood, so nothing is really lost.

banesullivan: Sounds good to me, thanks for handling this @AlmightyYakob!

banesullivan (approved):

matt-bernstein (approved):

matt-bernstein:
This is the failed test:

```python
def test_datastore_valid():
    """Validates that all entries in the datastore are available for download."""
    for k in datastore.registry.keys():
        assert datastore.datastore.is_available(k), f'`{k}` not available.'
```

Doesn't seem to have anything to do with this MR? Could just be a random 503 error. Maybe try rerunning the checks.

AlmightyYakob:
Rerunning the tests seem to have fixed it.

PR #365, opened 2021-04-23
Add settings based upload_to param for

banesullivan:
To be able to configure the upload_to path dynamically while also having a random uuid in the name to prevent overwriting files by accident

Needed for AWS deployment on another project where we need to put all S3FF objects under a specific "directory" in the S3 bucket


rgd/utility.py
```diff
@@ -240,3 +241,11 @@ def patch_internal_presign(f: FieldFile):
             f.storage.base_url = original_base_url
         return
     yield
+
+
+def uuid_prefix_filename(instance: Any, filename: str):
+    """Use a variable in settings to add a prefix to the path and keep the random uuid."""
+    prefix = getattr(settings, 'S3_FILE_FIELD_PREFIX', None)
```
brianhelba, line 248: I think you should name this setting differently, since it's part of RGD, not S3FF.

brianhelba, line 248: If you give this a default value of `None` in `settings.py`, you can rely on it always being defined, which might make maintenance easier in the future.

dchiquito, line 248:
```suggestion
    prefix = settings.S3_FILE_FIELD_PREFIX
```

I think it would probably be a good idea to add `None` as a default value in `settings.RgdMixin`:
```
    S3_FILE_FIELD_PREFIX = values.Value(default=None, environ=True)
```
That way all customized settings are explicitly called out in `settings.py`.

banesullivan: Fair point, how about `RGD_FILE_FIELD_PREFIX`?

banesullivan: Thanks for the tip!

rgd/settings.py
```diff
@@ -90,6 +90,12 @@ def before_binding(configuration: Type[ComposedConfiguration]):
 
     CELERY_WORKER_SEND_TASK_EVENTS = True
 
+    RGD_FILE_FIELD_PREFIX = values.Value(
+        default=None,
+        environ=True,
+        environ_prefix='DJANGO',
```
brianhelba, line 96: FYI, these are the default values of `environ` and `environ_prefix`. They typically do not deviate from these values, so it's surprising to see them explicitly set here. They could be omitted as arguments. See: https://django-configurations.readthedocs.io/en/latest/values/#configurations.values.Value

brianhelba: @banesullivan

banesullivan: Thanks for pointing this out. This was unclear to me and to be safe, I mimicked what I saw right above it: https://github.com/ResonantGeoData/ResonantGeoData/blob/42e6180d82d0e4e0052bfdf2e57b0667f7028f95/rgd/settings.py#L84

I have added #375 to clean these up

PR #368, opened 2021-04-23
Add new AWS production config

banesullivan:
Needed for AWS prod instance where AWS S3 credentials are managed by the AWS node that the instance is running on

### DIFF
--- a/rgd/settings.py
+++ b/rgd/settings.py
@@ -104,6 +104,12 @@ class ProductionConfiguration(RgdMixin, ProductionBaseConfiguration):
     pass
 
 
+class AWSProductionConfiguration(RgdMixin, ProductionBaseConfiguration):
+    AWS_S3_ACCESS_KEY_ID = None
+    AWS_S3_SECRET_ACCESS_KEY = None
+    AWS_S3_REGION_NAME = None
+
+
 class HerokuProductionConfiguration(RgdMixin, HerokuProductionBaseConfiguration):
     # Use different env var names (with no DJANGO_ prefix) for services that Heroku auto-injects
     DATABASES = values.DatabaseURLValue(


aashish24:
@banesullivan just curious do we set  AWSProductionConfiguration values  in some config file during the deployment?

banesullivan:
We normally set AWS credentials in environment variables. For the AWS production instance, all of this is managed by AWS and when we use `boto3` it all just works out of the box so we need to make sure our settings ignore these credentials.

banesullivan:
This also needs `SECURE_SSL_REDIRECT=False`

brianhelba:
> This also needs `SECURE_SSL_REDIRECT=False`

This shouldn't need to be the case (see [the note in these docs](https://docs.djangoproject.com/en/3.2/ref/settings/#secure-ssl-redirect)). Whichever proxy is doing SSL termination should be setting a header (often `HTTP_X_FORWARDED_PROTO`) to indicate the request protocol. This should be provided to the setting `SECURE_PROXY_SSL_HEADER` (see [the setting in composed-config](https://github.com/girder/django-composed-configuration/blob/c9006a5f0350e39ab2e51b424e22618d4f5ba28f/composed_configuration/_https.py#L12) and [the setting docs](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-SECURE_PROXY_SSL_HEADER)).

brianhelba:
> @banesullivan just curious do we set AWSProductionConfiguration values in some config file during the deployment?

@aashish24 Yes, the environment variable `DJANGO_CONFIGURATION` [must be set at runtime](https://github.com/ResonantGeoData/ResonantGeoData/blob/42e6180d82d0e4e0052bfdf2e57b0667f7028f95/rgd/wsgi.py#L7) (in this case, to the value `AWSProductionConfiguration`).

brianhelba:
@banesullivan What you have so far makes sense. Let me know if you want a final review once this is ready to merge.

brianhelba:
[The docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html) state
> Application Load Balancers ... automatically add `X-Forwarded-For`, `X-Forwarded-Proto`, and `X-Forwarded-Port` headers to the request.

`X-Forwarded-Proto` is what we want: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto

rgd/settings.py
```diff
@@ -104,6 +104,25 @@ class ProductionConfiguration(RgdMixin, ProductionBaseConfiguration):
...
+    AWS_S3_REGION_NAME = None
+    SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'http')
+
+    @staticmethod
+    def before_binding(configuration: Type[ComposedConfiguration]):
+        # Create a new file handler
+        configuration.LOGGING['handlers']['file'] = {
+            'level': 'NOTSET',
+            'class': 'logging.FileHandler',
+            'filename': '/web-apps/logs/rgd.log',
```
banesullivan, line 119: Should we use an environment variable to set this path?

brianhelba: https://12factor.net/logs
A more standard practice for 12-factor apps, even if they're running on EC2, is to log all output to stdout. The init system (supervisor, systemd, etc.) can then be responsible for routing output to log files (which are often stored under `/var/log` or `journalctl` with systemd) and handle tasks like log rotation.

An important reason you wouldn't want a webserver to write to a log file itself is concurrency. If you have multiple webserver processes (which is how a pre-fork WSGI server like Gunicorn scales), each will independently open and start writing to the same file simultaneously.

rgd/settings.py
```diff
@@ -104,6 +104,29 @@ class ProductionConfiguration(RgdMixin, ProductionBaseConfiguration):
     pass
 
 
+class AWSProductionConfiguration(RgdMixin, ProductionBaseConfiguration):
+    AWS_S3_ACCESS_KEY_ID = None
+    AWS_S3_SECRET_ACCESS_KEY = None
+    AWS_S3_REGION_NAME = None
+    SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'http')
```
brianhelba, line 111: This has the semantics that that "if `HTTP_X_FORWARDED_PROTO` has the value `http`, then the request is secure".

I don't know what the SSL-terminating proxy for this deployment is, but I would expect that this should be:
```suggestion
    SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
```

brianhelba (commented):
@banesullivan Reviewed. Comments inline.

rgd/settings.py
banesullivan: I'm not familiar with how this all works, but we are seeing infinite redirects when this is set to `https` on the AWS deployment. @chriswilley, can you provide any insight into the nginx setup for that deployment and what we are needing to set this to `http`? Perhaps, something with the nginx setup or load balancer needs to be changed?

chriswilley:
As you may recall, we ran into this a few weeks ago. Essentially, SSL communication from a client terminates at the application load balancer (ALB). From that point to the RGD application (within the VPC), traffic comes through using http. The ALB then encrypts the response back to the client. If this Django configuration setting is necessary, then I think it should be left as 'http'. Let me know if I'm missing anything.

banesullivan:
I set `SECURE_PROXY_SSL_HEADER` so that it can be managed by an ENV variable since we need it to be `http` for our specific deployment with the load balancer

Further, I removed the file logging handler and will set up the running service to pipe all stdout/stderror to a log file

rgd/settings.py
banesullivan: Thanks for the tip!

brianhelba:
> I set SECURE_PROXY_SSL_HEADER so that it can be managed by an ENV variable since we need it to be http for our specific deployment with the load balancer

@banesullivan I'm not clear what's going on here, but I'd recommend against the current configuration. Setting this value to `http` is likely "wrong", even if it's an easy hack to avoid errors.

Ideally, AWS ALB ought to be setting [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) properly (i.e. it should set `X-Forwarded-Proto: https` if it is performing SSL termination). If that can't be done, then I'd suggest disabling `SECURE_SSL_REDIRECT` and `SECURE_PROXY_SSL_HEADER` entirely.

PR #370, opened 2021-04-26
Add password prompt to Rgdc class constructor

AlmightyYakob:
This adds the ability to create an authenticated client instance, without needing to specify your username/password in plain text.

### DIFF
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -1,5 +1,6 @@
 from base64 import b64encode
 from dataclasses import dataclass
+import getpass
 import json
 from json.decoder import JSONDecodeError
 from pathlib import Path
@@ -28,6 +29,18 @@ class RasterDownload:
 
 
 class Rgdc:
+    @classmethod
+    def from_login_prompt(cls, *args, **kwargs):
+        """
+        Return an authenticated Rgdc instance from a login prompt.
+
+        Accepts the same positional/keyword arguments as the class constructor EXCEPT for `username` and `password`.
+        """
+        username = input('Username: ')
+        password = getpass.getpass()
+
+        return cls(*args, **kwargs, username=username, password=password)
+
     def __init__(
         self,
         api_url: str = DEFAULT_RGD_API,
--- a/rgdc/rgdc/rgdc.py
+++ b/rgdc/rgdc/rgdc.py
@@ -29,18 +29,6 @@ class RasterDownload:
 
 
 class Rgdc:
-    @classmethod
-    def from_login_prompt(cls, *args, **kwargs):
-        """
-        Return an authenticated Rgdc instance from a login prompt.
-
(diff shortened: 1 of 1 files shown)


aashish24:
this is great. It would be also good in the future if we can have some sort of tokens thoughts @matt-bernstein @AlmightyYakob @banesullivan

matt-bernstein (approved):
This does not allow a user to enter a username normally and then prompt for a password, as putting this logic in `Rgdc.__init__` would. However, I think this is ok, because you can always do something like:
`client = Rgdc(username='username', password=os.environ['RGD_PASS'])` instead.

AlmightyYakob:
> this is great. It would be also good in the future if we can have some sort of tokens thoughts @matt-bernstein @AlmightyYakob @banesullivan

I think it makes perfect sense, we could just add a similar function to this. I think it makes sense to also add support for environment variables, maybe something like `RGD_API_URL`, `RGD_USERNAME`, `RGD_PASSWORD`, `RGD_API_TOKEN`. However I'm not sure exactly where we'd fit them in / how they'd be used alongside the existing login schemes, but that's not an issue for now.

AlmightyYakob:
> This does not allow a user to enter a username normally and then prompt for a password, as putting this logic in `Rgdc.__init__` would. However, I think this is ok, because you can always do something like:
> `client = Rgdc(username='username', password=os.environ['RGD_PASS'])` instead.

I didn't realize this was the intention. I can change this to behave in this way, unless you think my initial implementation is better. Now that you mention it, I actually think including this in the class constructor would be a bit nicer.

aashish24:
> This does not allow a user to enter a username normally and then prompt for a password, as putting this logic in `Rgdc.__init__` would. However, I think this is ok, because you can always do something like:
> `client = Rgdc(username='username', password=os.environ['RGD_PASS'])` instead.

I thought this PR will allow you to enter password interactively, is that not the case?

matt-bernstein:
> I thought this PR will allow you to enter password interactively, is that not the case?
That is correct. However, right now you must enter the user and pw either both interactively, or both in code. Putting this in the constructor under a `if username is not None and password is None` would allow entering only the password interactively.

That might be marginally more convenient for the particular use case I have right now, but frankly I don't think it matters in the long run. I'm ok with this being merged as-is.

AlmightyYakob:
> How that would affect the current auth model - what username would you give to someone who passed in only an API key?

I don't think you'd login with a username and an API key, they'd be separate authentication schemes. This may require a change on the API. I'm not even sure if the API currently deals with API tokens at all.

> Would you make a separate API key for each user, or make a special `defaultuser` that cannot log in using a password? Not sure what the best practices are around this.

I think you'd have an API for each user, I don't know what the benefit of the latter would be.

AlmightyYakob:
@matt-bernstein I've updated this branch with a simple password prompt in the constructor. If that looks good then I think we're good to go here. The environment variable discussion should be a separate issue IMO.

PR #373, opened 2021-04-26
Add standalone workflow for tags/releases

AlmightyYakob:
With the previous setup, creating releases through Github wouldn't trigger the PyPI publish workflow.


banesullivan (approved):
Thanks!

PR #378, opened 2021-05-03
Attempt fix watch subdomain

banesullivan:
Fixing error on TF: https://app.terraform.io/app/ResonantGeoData/workspaces/ResonantGeoData/runs/run-o42QXWUxzrBAoVej

This is one of the zones for `us-west-2`: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-regions

### DIFF
--- a/terraform/watch.tf
+++ b/terraform/watch.tf
@@ -13,7 +13,7 @@ resource "aws_route53_record" "www-watch" {
 
   alias {
     name                   = "RDG-ELB1-758710195.us-west-2.elb.amazonaws.com"
-    zone_id                = "us-west-2-lax-1a"
+    zone_id                = "Z1H1FL5HABSF5"
     evaluate_target_health = true
   }
 }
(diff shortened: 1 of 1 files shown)


terraform/watch.tf
```diff
@@ -13,7 +13,7 @@ resource "aws_route53_record" "www-watch" {
 
   alias {
     name                   = "RDG-ELB1-758710195.us-west-2.elb.amazonaws.com"
-    zone_id                = "us-west-2"
+    zone_id                = "Z1H1FL5HABSF5"
```
brianhelba, line 16: If there's a way to obtain this which is less magical, it'd be great to wrap this in a Terraform data source. Alternatively, add some comments here explaining where this (and the `name` attribute) comes from.

Structurally, this looks more like a Route 53 Hosted Zone Id, so I think it's more likely to apply successfully.

brianhelba (approved):

terraform/watch.tf
banesullivan: The `name` and `zone_id` were given to me as read from the AWS management console for the record we are forwarding to.

I don't think we can wrap these values in a TF data source without taking over the management of this external ELB resource

brianhelba: By definition, a "data source" is a non-managed external entity. A "resource" would be something that's actively managed by TF. https://www.terraform.io/docs/language/data-sources/index.html

PR #382, opened 2021-05-05
Fix outline transform for rasters

### LINKED ISSUE #372: Support non-rectangular bounding box in SpatialEntry `footprint` or `outline`
`SpatialEntry.footprint` and `SpatialEntry.outline` seem to be identical in every `RasterMetaEntry` I have downloaded. I am not sure what the difference is supposed to be. The following is a use case that should be supported in whichever one of them `search` checks.

A query for `{'type':'Point','coordinates':[128.6643, 37.6601]}`, `instrumentation='OLI_TIRS'` (Landsat-8) in RGD looks like this:
image

But the Landsat path-row that covers that point (enter same coordinates into https://landsat.usgs.gov/landsat_acq#convertPathRow) looks like this:
image

because the Landsat WRS-2 coordinate system is offset from the standard projection (EPSG 4326?) that RGD uses.

It seems that during ingestion into RGD, the EPSG4326/WGS84 bounding box of the scene is used as `footprint` and `outline`. But the true outline can be found in the files:

```bash
$ grep 'LAT_PRODUCT\|LON_PRODUCT' LE07_L1TP_115034_20181102_20181129_01_T1/LE07_L1TP_115034_20181102_20181129_01_T1_MTL.txt
    CORNER_UL_LAT_PRODUCT = 38.43903
    CORNER_UL_LON_PRODUCT = 126.71531
    CORNER_UR_LAT_PRODUCT = 38.46006
    CORNER_UR_LON_PRODUCT = 129.54100
    CORNER_LL_LAT_PRODUCT = 36.49358
    CORNER_LL_LON_PRODUCT = 126.77379
    CORNER_LR_LAT_PRODUCT = 36.51319
    CORNER_LR_LON_PRODUCT = 129.52714
```
(There are probably more reliable ways to do this... I will dig around some more. [This](https://gis.stackexchange.com/questions/372160/issue-while-creating-dataset-mask-for-landsat-7-etm-band-data-by-using-rasterio) could be a good starting point.)

The reason this is a problem is that it could cause false positives in a search for those triangular spatial regions outside the path-row but inside the bbox. However, this is not a fundamental flaw because a user who knows to look for a `nodata` mask can correct for this after downloading the images.

So there are several action items here:
1. Allow nonrectangular bounding box in `SpatialEntry` - does any work need to be done for this?
2. For Landsat specifically, read the valid data mask or the path-row in the shapefile to get the true footprint
3. For all satellites, handle reboxing correctly when converting from source CRS to RGD's CRS (in which the rectilinear bbox from the source CRS becomes diagonal)

banesullivan:
This fixes an issue when computing the outline of rasters. Previously, the transform was done incorrectly and would lead to errors in larger images. Here is an example with Landsat imagery:

| `master` | this branch |
|---|---|
|  Screen Shot 2021-05-05 at 8 53 54 AM |  Screen Shot 2021-05-05 at 9 56 15 AM |

Further, I am working to resolve an issue when computing the valid data mask (`footprint`) when the `nodata` is set to `None`

resolves #372

### DIFF
--- a/rgd/geodata/management/commands/_data_helper.py
+++ b/rgd/geodata/management/commands/_data_helper.py
@@ -13,6 +13,7 @@ from django.utils.timezone import make_aware
 
 from rgd.geodata import models
 from rgd.geodata.datastore import datastore
+from rgd.geodata.models.imagery import etl
 from rgd.utility import get_or_create_no_commit
 
 logger = logging.getLogger(__name__)
@@ -128,7 +129,7 @@ def load_image_files(image_files):
     return ids
 
 
-def load_raster(pks, raster_dict):
+def load_raster(pks, raster_dict, footprint=False):
     if not isinstance(pks, (list, tuple)):
         pks = [
             pks,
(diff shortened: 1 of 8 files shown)


aashish24:
the improvements are great, very  noticeable

banesullivan:
This further addresses an issue when the `nodata` value on a raster is not set for computing the valid data footprint (such is the case for Landsat imagery). This now can handle computing the footprints properly (red outline):

Screen Shot 2021-05-05 at 11 48 24 AM

This means results can be greatly improved when searching to address #372 

I have also added a new option for all of the management commands to compute the valid data footprint on ingest if desired since it is an optional feature.

PR #396, opened 2021-05-11
Add new SpatialAsset model and change footprint field

### LINKED ISSUE #328: Annotated Geo Asset Model
Add a new model for arbitrary assets that have an annotated spatial context.

e.g. a 3rd party has a PowerPoint presentation and a data scrubbing routine to extract spatial metadata from that PPT. They would then populate this new annotated geo asset model with the extracted spatial metadata and point to it from another custom table holding their PPT or other random data formats. Then these entries would be searchable in a geospatial context.

think of all the different types of data that are not traditionally geospatial but could have geospatial context extracted or assigned: documents, presentations, etc.

banesullivan:
Resolve #328 

Adds a new `SpatialAsset` model for making a `SpatialEntry` out of any collection of files not explicitly handled by existing models (`RasterEntry`/`RasterMetaEntry`, `FMVEntry`, and `GeometryEntry`).

This also changes the `footprint` field on `SpatialEntry` to be a `GeometryField` rather than a `PolygonField` such that we can use arbitrary geometry as the spatial "footprint". This is to cover a use case where a `SpatialAsset` might only be associated with a single point or a line rather than a polygon.


banesullivan:
These changes are required for https://github.com/ResonantGeoData/issue-dashboard/issues/12 and https://github.com/ResonantGeoData/issue-dashboard/issues/13

mcovalt (approved):

PR #397, opened 2021-05-11
Add new heatmap stats view

### LINKED ISSUE #389: Heat map to show all data in the catalog
We should add a new heatmap page to show a heat map of all of the data in the catalog.

It would be nice to also show this heatmap during the search, but I'm not sure how best to handle that since we do want the footprints of the paginated search results to show up too

banesullivan:
Resolve #389

Screen Shot 2021-05-11 at 9 29 20 AM

To do:

- [x] Test against https://github.com/OpenGeoscience/geojs/pull/1088
- [x] Add more fields to the column on the left
- [x] Set default map extents on page load
- [ ] Make sure this works with large amounts of data (should we make the centroid fetching into an async endpoint?)


rgd/geodata/templates/geodata/statistics.html
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+
+    <div class="column p-0">
+      {% include 'geodata/_includes/empty_viewer.html' %}
+
+      <script>
+        var mapviews = document.getElementsByClassName('geojs-map');
+        for (var i = 0; i < mapviews.length; i++) {
+          mapviews[i].style.overflow = 'hidden';
+        }
```
banesullivan, line 60: Temporary fix until https://github.com/OpenGeoscience/geojs/pull/1088 is released

PR #401, opened 2021-05-17
Add setting to allow access to unassigned permissions for all users

banesullivan:
Adds a new setting to give global read access to records that do not have assigned permissions. 

This is useful if we want all users to have access to all data in an instance by default and still be able to make certain collections private.

### DIFF
--- a/rgd/geodata/permissions.py
+++ b/rgd/geodata/permissions.py
@@ -105,7 +105,7 @@ def filter_perm(user, queryset, role):
     filtered = queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
     # Check setting for unassigned permissions
     if settings.RGD_GLOBAL_READ_ACCESS:
-        unassigned = queryset.filter(**{'_collection_permissions__user': None})
+        unassigned = queryset.filter(**{user_path: None})
         return unassigned.union(filtered)  # This union throws an error
     return filtered
 
--- a/rgd/geodata/permissions.py
+++ b/rgd/geodata/permissions.py
@@ -1,5 +1,6 @@
 from typing import Optional
 
+from django.conf import settings
 from django.contrib.auth.backends import BaseBackend
 from django.core.exceptions import PermissionDenied
 from django.db.models.functions import Coalesce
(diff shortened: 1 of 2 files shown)


rgd/settings.py
```diff
@@ -14,6 +14,9 @@
 
 
 class GeoDjangoMixin(ConfigMixin):
+    # A setting to allow read access to all users if permissions are unassigned
+    RGD_GLOBAL_READ_ACCESS = values.Value(default=True)
```
banesullivan, line 18: Need to set to `False` as default. `True` for now for dev

rgd/geodata/permissions.py
```diff
@@ -86,22 +87,27 @@ def filter_perm(user, queryset, role):
...
     # `path` can be an empty string (meaning queryset is `CollectionPermission`)
     user_path = (path + '__' if path != '' else path) + 'user'
     role_path = (path + '__' if path != '' else path) + 'role'
     queryset = annotate_queryset(queryset)
-    return queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
+    filtered = queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
+    # Check setting for unassigned permissions
+    if settings.RGD_GLOBAL_READ_ACCESS:
+        unassigned = queryset.filter(**{'_collection_permissions__user': None})
+        return unassigned.union(filtered)  # This union throws an error
```
banesullivan, line 109: @mcovalt, can you help get the logic/filtering right here? This union is throwing an error at the DB level

rgd/geodata/permissions.py
```diff
@@ -86,26 +87,31 @@ def filter_perm(user, queryset, role):
...
     # `path` can be an empty string (meaning queryset is `CollectionPermission`)
     user_path = (path + '__' if path != '' else path) + 'user'
     role_path = (path + '__' if path != '' else path) + 'role'
     queryset = annotate_queryset(queryset)
-    return queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
+    filtered = queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
+    # Check setting for unassigned permissions
+    if settings.RGD_GLOBAL_READ_ACCESS:
+        unassigned = queryset.filter(**{user_path + '__isnull': True})
+        return unassigned | filtered
```
banesullivan, line 109: This does not use SQL to perform a union, but joins the querysets after they have been fetched

rgd/geodata/permissions.py
banesullivan: Fixed, see other comment

rgd/geodata/permissions.py
banesullivan: I was seeing an error when trying to do the union via SQL/the queryset's `union()` method

mcovalt (approved):

PR #402, opened 2021-05-18
Refactor serializers and add new STAC API view

### LINKED ISSUE #352: Clean up endpoints and RGDC client
- [x] Make separate search endpoints for each spatial type
- [x] Clean up "_entry" occurrences in Python client, etc.

### LINKED ISSUE #374: `as_stac` option for `Rgdc.search`
Before https://github.com/ResonantGeoData/ResonantGeoData/issues/327 is complete, one thing we can do in the meantime is add an option for `Rgdc.search` results to conform to the fields of a [STAC item](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields). This will be useful for downstream tasks that expect to ingest a STAC catalog.

In this case, there will not be a true catalog, but one can be created based on known information in the search result (`instrumentation`, tile, date).

banesullivan:
- Refactoring of filters to separate the SpatialEntry fields from domain model fields
    - removes the ModifiableEntry fields from the filters (there is a new `ModifiableEntryFilterMixin` but it is unused because we can't really mix filters)
- Utilizes STAC Item serializer from #386 for new search endpoints
    - returns raster search results in STAC Item 
- Updates front-end templates to use new search views/filters

Resolve #352, resolve #374, resolve https://github.com/ResonantGeoData/issue-dashboard/issues/10


banesullivan:
@AlmightyYakob, would you be able to review the changes to the Python client in `rgdc/rgdc/rgdc.py` with these new endpoints/changes in existing endpoints?

rgdc/rgdc/rgdc.py
```diff
@@ -206,19 +206,85 @@ def search(
         relates: Optional[str] = None,
         distance: Optional[Tuple[float, float]] = None,
         acquired: Optional[DATETIME_OR_STR_TUPLE] = None,
-        created: Optional[DATETIME_OR_STR_TUPLE] = None,
-        modified: Optional[DATETIME_OR_STR_TUPLE] = None,
-        datatype: Optional[SEARCH_DATATYPE_CHOICE] = None,
```
AlmightyYakob, line 211: If `datatype` won't be used anymore we should remove `SEARCH_DATATYPE_CHOICE` from `types.py`

rgdc/rgdc/rgdc.py
```diff
@@ -206,19 +206,85 @@ def search(
...
+
+        # TODO: Determine if the before/after param order needs to be swapped?
+        if acquired and len(acquired) == 2:
+            amin, amax = acquired
+            params['acquired_before'] = datetime_to_str(amax)
+            params['acquired_after'] = datetime_to_str(amin)
+
+        return list(limit_offset_pager(self.session, 'geosearch', params=params))
+
+    def search_raster_stac(
```
AlmightyYakob, line 272: I'm trying to think about this in a similar way to the `get_raster` method, for which we combined the two methods into one and used the `stac` parameter to specify the difference. This is obviously more complex so I'm not sure it will apply in quite the same way, but there is a good deal of duplicated code between this and the regular search endpoint. @banesullivan Do you think it's worthwhile to try to bridge the gap between these two methods? I'm okay with keeping them completely separate, at least for now.

rgdc/rgdc/rgdc.py
banesullivan: Done

rgdc/rgdc/rgdc.py
banesullivan: I added a new `_spatial_search_params` to build the params for the SpatialEntry search which both methods can use to reduce duplicate code.

I do not want to combine these two methods because they are operating on different data types (even when we had the datatype param, we were only operating on SpatialEntry records).

banesullivan: Done

AlmightyYakob: Did you push your changes? I don't see anything new

rgdc/rgdc/rgdc.py
```diff
@@ -206,19 +203,59 @@ def search(
...
+            distance=distance,
+            acquired=acquired,
+            instrumentation=instrumentation,
+            limit=limit,
+            offset=offset,
+        )
+        return list(limit_offset_pager(self.session, 'geosearch', params=params))
+
+    def search_raster_stac(
+        self,
```
AlmightyYakob, line 244: One last thing, this param needs to be removed
```suggestion
```

rgdc/rgdc/utils.py
```diff
@@ -122,3 +127,53 @@ def download_checksum_file_to_path(file: Dict, path: Path, keep_existing: bool):
 def spatial_subentry_id(search_result):
     """Get the id of a returned SpatialEntry."""
     return search_result['subentry_pk']
+
+
+def spatial_search_params(
+    self,
```
AlmightyYakob, line 133: One last thing, this parameter needs to be removed
```suggestion
```

banesullivan: Thanks!

AlmightyYakob (approved):

mcovalt (approved):

Issue #307, opened 2021-03-11
large_image tiling issue with specific dataset


banesullivan:
Recently, `large_image` started failing to serve tiles for this dataset: https://data.kitware.com/#item/5f9c316650a41e3d1920ef90

We were originally using this raster for some testing of the `large_image` tile serving endpoint, so I know it worked at one point.

Recently, I started seeing this `numpy` error when trying to serve tiles for this file:

```
...
                                  File "/usr/local/lib/python3.8/site-packages/nu
django_1    |                     mpy/core/getlimits.py", line 523, in __init__
django_1    |                         raise ValueError("Invalid integer data type
django_1    |                     %r." % (self.kind,))
django_1    |                     ValueError: Invalid integer data type 'f'.
```

<details>

```
django_1    | [16:42:08] ERROR    Internal Server Error: /api/geodata/imagery/image log.py:224
django_1    |                     _entry/159/tiles/12/830/1565.png
django_1    |                     Traceback (most recent call last):
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/core/handlers/exception.py", line 47, in
django_1    |                     inner
django_1    |                         response = get_response(request)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/core/handlers/base.py", line 181, in
django_1    |                     _get_response
django_1    |                         response = wrapped_callback(request,
django_1    |                     *callback_args, **callback_kwargs)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/views/decorators/csrf.py", line 54, in
django_1    |                     wrapped_view
django_1    |                         return view_func(*args, **kwargs)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/views/generic/base.py", line 70, in view
django_1    |                         return self.dispatch(request, *args,
django_1    |                     **kwargs)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/re
django_1    |                     st_framework/views.py", line 509, in dispatch
django_1    |                         response = self.handle_exception(exc)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/re
django_1    |                     st_framework/views.py", line 469, in
django_1    |                     handle_exception
django_1    |                         self.raise_uncaught_exception(exc)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/re
[...]

PR #403, opened 2021-05-18
Set large_image version

banesullivan:
Follow up to https://github.com/ResonantGeoData/ResonantGeoData/issues/307#issuecomment-797640849

We are seeing #307 again on the Heroku instance lately. Hopefully setting the min version will prevent an older large_image from being installed

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
         'djangorestframework',
         'drf-yasg',
         'GDAL',
-        'large-image>=1.4.3',
+        'large-image>=1.6.0',
         'large-image-source-gdal',
         'numpy',
         'pooch',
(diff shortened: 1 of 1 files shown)


banesullivan:
@manthey, should we bump to `1.5.x` to also utilize https://github.com/girder/large_image/pull/594 ?

manthey:
> @manthey, should we bump to `1.5.x` to also utilize [girder/large_image#594](https://github.com/girder/large_image/pull/594) ?

We haven't cut a release with https://github.com/girder/large_image/pull/594, so currently it would have to be pip installed with the `--pre` flag.  Probably the right thing to do is to make a large_image release and then use that version as the minimum.

manthey (approved):

PR #406, opened 2021-05-20
Point Clouds and new 3D VTK.js viewer

banesullivan:
Adds a new set of models for ingesting and storing point clouds.

Resolves https://github.com/ResonantGeoData/issue-dashboard/issues/12

This handles importing a wide variety of 3D point clouds using PyntCloud and PyVista. This PR currently contains a minimal, first pass of front-end detail view for the new PointCloudEntry model using VTK.js

The `PointCloudEntry` model is optionally enabled to be listed as a `SpatialEntry` via the `PointCloudMetaEntry`. This means users can upload point clouds to the catalog that have no spatial reference. At present, there is no browser for these data other than via the admin interface. 

`PointCloudEntry` records can be manually created with VTP data without an associated `PointCloudFile` though the best approach is to always upload via `PointCloudFile`. We could make that relationship non-nullable but I haven't here for flexibility.

The `PointCloudEntry` stores a copy of the data uploaded in `PointCloudFile` as a converted VTP file ready for VTK.js

This work adds a few not-so-light dependencies to the worker:
- VTK and PyVista (in a headless mode) for converting to the VTP format (necessary to view in VTK.js)
- PyntCloud, `pylas`, and `lazrs` for loading LAS point cloud files

## Preview

Screen Shot 2021-05-23 at 5 39 33 PM

## To do

- [x] Fix off-center VTK.js view
- [x] Add permissions for PointCloudFile and other models
- [x] Create an endpoint for accessing point cloud data and have the front-end view asynchronously load the data rather than sending the data through the page build context (maybe add a "loading..." icon over the view)
- [x] Improve metadata column on the left side of detail view (add download links, file size, spatial context if it has it, etc.)
- [x] Find permissive data for testing and examples
- [x] Doubly check new VTK dependency for a headless worker (VTK+PyVista provides the most/easiest file conversions)
- [x] Make sure `PointCloudMetaEntry` shows up as a subtype and is listed in search view


rgd/geodata/models/threed/etl.py
```diff
@@ -0,0 +1,80 @@
...
+        with open(output_path, 'rb') as f:
+            output_field.save(os.path.basename(output_path), f)
+
+
+def _save_pyvista(mesh, output_path):
+    import pyvista as pv
+
+    points = pv.PolyData(mesh.points)
+    points.point_arrays.update(mesh.point_arrays)
+    points.save(output_path)
```
banesullivan, line 30: NOTE: this extracts ONLY the points of the mesh and the point data arrays. We could change this to save any geometry that mesh has as well.

For example, if a user uploads a triangulated mesh, they will only see the point vertices in the front-end viewer.

If we want to always save a PolyData mesh of points and geometry, then we could change this to:

```suggestion
    mesh = mesh.extract_surface()
    mesh.save(output_path)
```

banesullivan:
Note that the filter search might not work on this... This will be fixed by #402

aashish24:
Nice going @banesullivan this is great. I am assuming that this point cloud does not have color information yet?

banesullivan:
> I am assuming that this point cloud does not have color information yet?

In this PR, no -- this is not supported yet.

I'm hoping to improve the viewer in follow up work which is captured in #407

aashish24:
Okay thanks so I took it as that the data has the color information but we are not supporting it in this PR. Thanks.

PR #409, opened 2021-05-24
Add default site migration

### LINKED ISSUE #388: Remove "example.com" in Signup Confirmation Email
The signup email lists `example.com` all over the place. We need to fix this: (I removed my email and the signup link which were correct)

```
Hello from example.com!

You're receiving this e-mail because user ..... has given your e-mail address to register an account on example.com.

To confirm this is correct, go to ........

Thank you for using example.com!
example.com
```

-----

It seems that `site_name` in https://github.com/girder/django-girder-style/blob/master/girder_style/templates/account/email/base_message.txt is incorrect

banesullivan:
Resolve #388 

Somewhere along the way, this migration was deleted and not re-added

### DIFF
--- a/rgd/geodata/migrations/0016_default_site.py
+++ b/rgd/geodata/migrations/0016_default_site.py
@@ -1,0 +1,37 @@
+from django.conf import settings
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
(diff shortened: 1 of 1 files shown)


banesullivan:
Closing since the prod DB already has this reset. I will keep #388 open to remind us down the road if we ever squash the migrations, to re-add

PR #411, opened 2021-05-25
Fix duplicate search results from permissions

### LINKED ISSUE #410: Unassigned permissions not working
Non-admin accounts are seeing an incorrect amount of spatial entries in the search results. Here are two results: 1) from my `@kitwar`e account which is an admin and one from my `@gmail` account that has no permissions. Using the changes from #401: 

https://github.com/ResonantGeoData/ResonantGeoData/blob/014ce2693a0a3e899d6af0a9d7822a5f1327268c/rgd/geodata/permissions.py#L108

You can see 475 results with the admin account (which is the correct amount) and 4949 results with the nonadmin account which hits that new code (this number is wildly incorrect): 

| admin | nonadmin |
| --- | --- |
| Screen Shot 2021-05-25 at 8 14 00 AM | Screen Shot 2021-05-25 at 8 13 51 AM |

banesullivan:
This resolves #410 by adding a `distinct()` call to the filterset and also adds a new test to verify all is working.

Further, I noticed we were re-using a lot of the same data fixtures in different testing modules so I went ahead and moved many of them to a new `data_fixtures` module

### DIFF
--- a/rgd/geodata/permissions.py
+++ b/rgd/geodata/permissions.py
@@ -102,10 +102,10 @@ def filter_perm(user, queryset, role):
     user_path = (path + '__' if path != '' else path) + 'user'
     role_path = (path + '__' if path != '' else path) + 'role'
     queryset = annotate_queryset(queryset)
-    filtered = queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role})
+    filtered = queryset.filter(**{user_path: user.pk}).exclude(**{role_path + '__lt': role}).distinct()
     # Check setting for unassigned permissions
     if settings.RGD_GLOBAL_READ_ACCESS:
-        unassigned = queryset.filter(**{user_path + '__isnull': True})
+        unassigned = queryset.filter(**{user_path + '__isnull': True}).distinct()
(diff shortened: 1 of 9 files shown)


banesullivan:
The failing test:

```py
    @pytest.mark.django_db(transaction=True)
    def test_unassigned_permissions_complex(user_factory, user, sample_raster_a, multi_file_raster):
        prior = settings.RGD_GLOBAL_READ_ACCESS
        admin = user_factory(is_superuser=True)
        admin_q = filter_read_perm(admin, models.RasterMetaEntry.objects.all())
        basic_q = filter_read_perm(user, models.RasterMetaEntry.objects.all())
        assert len(admin_q) == 2
        assert len(basic_q) == 0
        settings.RGD_GLOBAL_READ_ACCESS = True
        admin_q = filter_read_perm(admin, models.RasterMetaEntry.objects.all())
        basic_q = filter_read_perm(user, models.RasterMetaEntry.objects.all())
>       assert len(admin_q) == len(basic_q)
E       assert 2 == 4
E         +2
E         -4
```

PR #412, opened 2021-05-26
Performance Improvements

banesullivan:
Some major performance improvements and remove all of the old/unused search endpoints


rgd/geodata/permissions.py
```diff
@@ -108,7 +108,7 @@ def filter_perm(user, queryset, role):
     # Check setting for unassigned permissions
     if settings.RGD_GLOBAL_READ_ACCESS:
         unassigned = queryset.filter(**{user_path + '__isnull': True}).distinct()
-        return unassigned | filtered
```
banesullivan, line 111: this was previously evaluating the querysets and returning all records. We want this to remain as a queryset and let the views handle pagination

rgd/geodata/serializers.py
```diff
@@ -14,7 +14,7 @@
 class SpatialEntrySerializer(serializers.ModelSerializer):
     def to_representation(self, value):
         ret = super().to_representation(value)
-        ret['footprint'] = json.loads(value.footprint.geojson)
+        # NOTE: including footprint can cause the search results to blow up in size
```
banesullivan, line 17: NOTE: the default serializer no longer includes the footprint. There is a separate serializer and new endpoint to get it as some footprints can be very large

PR #416, opened 2021-06-01
Add STAC ingestion on serializer

mcovalt:
There's plenty of information not being grabbed, but this adds a routine to create models from STAC input. I anticipate this can be refined to grab more information as we get feedback from users.

- [x] Disable ETL on create
- [x] Expose API endpoint
- [x] Test against sample data


banesullivan (changes requested):
I thought a STAC Item could have multiple `assets`? In this configuration, if only grabs one data asset and creates a single raster for it. Don't we want to be able to grab multiple `data` assets and ancillary files for creating the single raster entry?

Ideally, we should be able to back import what we export with this Serializer

banesullivan:
This is pretty much done at this point.... it honestly feels hacky but it works

Caveats:

- this now requires `rasterio` and `shapely` as depends in the main web instance
- The exposed POST endpoint for this serializer runs some ETL routines the main request thread which could time out on some servers (We will deal with this when refactoring the models for https://github.com/ResonantGeoData/issue-dashboard/issues/17 as I will change the relationships to not depend on ETL-creaed models)
- pystac is telling us that our STAC JSON does not validate against the schema -- I will open a follow up issue about this. See commit df7fc2974e35adac0acf5da91ebcac5fbc6651d3

banesullivan:
Tested against sample data and this seems to be working fine.... merging.

PR #417, opened 2021-06-02
Fix Convex Hull of valid data footprint

banesullivan:
Whoops 🙃 

This makes sure we compute a reasonable polygon -- especially in the case of Landsat 7 ETM images with the scan line error

| before | after |
|---|---|
| Screen Shot 2021-06-02 at 9 33 10 AM | Screen Shot 2021-06-02 at 9 35 24 AM |


aashish24:
This looks great. Did it solve our issue or is there more that we need to get done?

banesullivan:
This solved the issue -- I just need to reprocess the Landsat 7 imagery on the AWS deployment which is already in progress (perhaps compete at this point)

PR #426, opened 2021-06-09
Improve query performance

### LINKED ISSUE #413: Permissions Queries Time Out
The `annotate_queryset` method annotates `SpatialEntry` querysets in such a way that we end up with massive queries to decide permissions for a given user.

For instance, let's imagine that all of our RasterEntries in the database each have 10 images (each from another file). This means that the following line will yield the same raster 10 times (factored by the number of users in the collections for each of those files):

https://github.com/ResonantGeoData/ResonantGeoData/blob/aaa11dd1ef6c6667a4cd37e15a868369552c506a/rgd/geodata/permissions.py#L23

Because of this, requests to the home page to load the full catalog for a non-admin user time out on our deployments.

Using the django debug toolbar, you can see some major differences in the SQL queries needed for an admin vs a regular user:

| admin | regular user |
|---|---|
| Screen Shot 2021-05-26 at 5 34 41 PM | Screen Shot 2021-05-26 at 5 34 53 PM |

The time for my instance running locally is relatively fast, but on actuall deployments, this time blows up and times out the request.

@mcovalt, I will need your help in finding a faster way to do these queries

mcovalt:
Closes #413

- The performance of general querying for permissions is improved. The joins in Django were `LEFT OUTER JOINS`. This caused PostgreSQL to choke when planning some queries. This new approach results in unions of `INNER JOINS`. Queries are reduced to <20 ms as opposed to +1,000 ms in my tests.
- Some effort was spent consolidating queries on the loading page. In some cases multiple queries were run per object. This work further decreased front page load time by a factor of 4.


banesullivan (approved):
Awesome work, @mcovalt!! This makes the AWS deployment super smooth/rapid

PR #431, opened 2021-06-09
More large_image utilization

banesullivan:
Resolve https://github.com/ResonantGeoData/issue-dashboard/issues/6


banesullivan:
getRegion with world coordinates is not working. Not sure why, but no output is produced. See failing tests.

Further the output when using getRegion with pixel coordinates on a non-geospatial JPEG image is not valid.

rgd/geodata/models/imagery/large_image_utilities.py
```diff
@@ -0,0 +1,38 @@
...
+    tile_source: FileTileSource,
+    left: float,
+    right: float,
+    bottom: float,
+    top: float,
+    projection: str = 'EPSG:3857',
+):
+    region = dict(left=left, right=right, bottom=bottom, top=top, units=projection)
+    path, mime_type = tile_source.getRegion(region=region, encoding='TILED')
+    return path, mime_type
```
banesullivan, line 30: @manthey, do you see anything immediately wrong with this? I cannot get it to produce an output when using world coordinates.

Example script outside of the django app:

```py
from large_image_source_gdal import GDALFileTileSource
from large_image.tilesource import FileTileSource

projection = 'EPSG:3857'
file_path = '/Users/bane/Downloads/Elevation.tif'
tile_source = GDALFileTileSource(file_path, projection=projection, encoding='PNG')


def get_region_world(
    tile_source: FileTileSource,
    left: float,
    right: float,
    bottom: float,
    top: float,
    projection: str = 'EPSG:3857',
):
    region = dict(left=left, right=right, bottom=bottom, top=top, units=projection)
    path, mime_type = tile_source.getRegion(region=region, encoding='TILED')
    return path, mime_type


bounds = tile_source.getBounds()

extent = -107.16011512365533, -107.05522782296597, 38.87471016725091, 38.92317443621267

assert extent[0] > bounds['xmin']
assert extent[1] < bounds['xmax']
assert extent[2] > bounds['ymin']
assert extent[3] < bounds['ymax']

path, _ = get_region_world(tile_source, *extent)

assert path, 'No output generated.'
with open(path, 'rb') as f:
    assert f.read()
```

Note that this leads to an assertion error:

```bash
$ python test_subsample.py
Created LRU Cache for 'tilesource' with 81 maximum size
Using python for large_image caching
Traceback (most recent call last):
  File "test_subsample.py", line 33, in <module>
    assert path, 'No output generated.'
AssertionError: No output generated.
```

`Elevation.tif`: https://data.kitware.com/#item/5f9c316650a41e3d1920ef90

banesullivan: Oh wait, this seems to just be a projection issue... the `region` is not in the right projection it appears.

banesullivan: Issue was simply the region's projection. Needed to be changed to `EPSG:4326`

PR #435, opened 2021-06-11
Refactor to series of Django Apps

### LINKED ISSUE #394: DateTime field in HTML form not formatting properly
The DateTime field on the HTML search form is broken (at least for the acquisition field). I can search against the acquisition field just fine via the Python client, but the format of the DateTime from the search form is different and not returning any results.

### LINKED ISSUE #405: Publish rgd to PyPI
We are beginning to use `rgd` as an installed app in other django/girder-4 apps and it would make dev way easier if we published `rgd` to PyPI on a regular basis

### LINKED ISSUE #419: Make sure all API views use get_object_or_404
If the object doesn't exist, we get server errors like: https://sentry.io/organizations/kitware-data/issues/2299463062/?project=5458973&query=is%3Aunresolved&statsPeriod=90d

This is from:

https://github.com/ResonantGeoData/ResonantGeoData/blob/86e049b87c3c6c58e2f74a2d05ef8c63acbee255/rgd/geodata/api/download.py#L28

Which should use `get_object_or_404` much like

https://github.com/ResonantGeoData/ResonantGeoData/blob/86e049b87c3c6c58e2f74a2d05ef8c63acbee255/rgd/geodata/api/tiles.py#L16

### LINKED ISSUE #420: Remove assertions due to unhandled server errors
Sentry error: https://sentry.io/organizations/kitware-data/issues/2366462093/?project=5458973&query=is%3Aunresolved&statsPeriod=90d

From this assertion:

https://github.com/ResonantGeoData/ResonantGeoData/blob/5370164cd2a9a3cac5451fbe31dab3467e02ae51/rgd/geodata/views.py#L41

Further, we should remove other assertions not in testing modules:

https://github.com/ResonantGeoData/ResonantGeoData/blob/5adc362c5367352b6e39c14a194dbb67874e6dd1/rgd/geodata/models/fmv/etl.py#L96

as it's bad practice to use assert in the codebase outside of the tests

### LINKED ISSUE #433: Make Geometry a non-required field in search form
We should be able to search against facets without the geometry.

e.g. find all data between 2 dates

banesullivan:
For https://github.com/ResonantGeoData/issue-dashboard/issues/17

- resolve #433
- resolve #420
- resolve #419 
- resolve #394 
- resolve #405

This PR completely resets migrations and splits out the `geodata` app into 5 new apps see https://github.com/ResonantGeoData/issue-dashboard/issues/17 for overview. Packages are currently published on PyPI as release candidates.

This further refactors the models such that there are no relationships that depend on ETL routines. Previously, trying to create an `ImageSet` after creating `ImageFile`s would require you to wait until the images were processed. Now, relationships are such that you can create `Image`s and the `ImageSet` points to those and all the processing occurs asynchronously. This was extended elsewhere.

rgd


banesullivan:
@brianhelba, would you please review the general structure, packaging, and CI implemented here? Things I'd appreciate feedback / a 2nd pair of eyes on:

- Each of the `setup.py` files for the new Apps - note that they are currently released on PyPI 🎉 
- The versioning system
    - there is one `version.py` file at the root of the repository. This is opened and `exec`'d in the `setup.py` file of each app. Then the core `django-rgd` app uses `pkg_resources.get_distribution` to get that version on import and the sub-packages import that same version from the core module (e.g. `from rgd import __version__` in their `__init__.py`)
- The testing setup with tox
- The CI matrix for testing in `.github/workflows/ci.yml`
- We may need to trim some of the dependencies in `django-rgd`'s `setup.py` as a lot of those were copied from the original Project/Cookiecuter

----

@mcovalt, would you be able to review/look over the implementation of the new permissions check? It's basically the same as you previously implemented, though now we have:

- a new interface for models that we have permissions defined through the `PermissionPathMixin` class in `django-rgd/rgd/models/mixins.py`
- in the `django-rgd` core app, I added a new way of "discovering" subclasses of `SpatialEntry` to get their permissions paths such that the core app does not need to have awareness of downstream apps that subclass `SpatialEntry` in order to filter the permissions. See `get_subclasses` and its use in `get_collection_permissions_paths` in `django-rgd/rgd/permissions.py`. Essentially, I find all of the subclasses of `SpatialEntry`, iterate over them, and aggregate their permissions paths before executing the query.
- Then since we are using the `InheritanceManager` in `SpatialEntry`, we can leverage `select_subclasses` to select all subclasses and return those in `filter_perm`

banesullivan:
merging...

PR #438, opened 2021-06-25
Add Terraform to store watch_dev state

brianhelba:
These resources were created manually, but have been imported into the TF state to ensure they get cleaned up.

### DIFF
--- a/terraform/watch.tf
+++ b/terraform/watch.tf
@@ -18,12 +18,12 @@ module "watch_smtp" {
   source  = "girder/girder/aws//modules/smtp"
   version = "0.8.0"
 
-  fqdn = aws_route53_record.watch.fqdn
-  project_slug = "watch"
+  fqdn            = aws_route53_record.watch.fqdn
+  project_slug    = "watch"
   route53_zone_id = aws_route53_zone.common.zone_id
 }
 
 output "watch_smtp_url" {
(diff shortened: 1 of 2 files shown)


banesullivan:
I'm actually going to create a new `RD-WATCH` repo for the WATCH deployment soon here as #435 is soon to merge and the main instance we have on Heroku is now under `RD-OpenGeo`: https://github.com/ResonantGeoData/RD-OpenGeo

see https://github.com/ResonantGeoData/RD-OpenGeo/issues/10

brianhelba:
@banesullivan As soon as you create the new repo for WATCH, @venkatabhishek can start to test deployment using that as a base Django project.

All the Terraform in this PR is ephemeral, and will be removed after WATCH deployment development ends.

More broadly, it will be much easier if we can keep all the TF for resources on the `RD-OpenGeo` AWS account under a single repo.  To the extent we want a DNS entry pointing to WATCH under the `resonantgeodata.com` domain, we'll need to keep that `aws_route53_record` with the rest of the TF. If we're looking for real separation, then maybe it's time for WATCH to register its own domain name.

It probably will be good for WATCH to get its own SMTP server too (which the WATCH support team should be able to help with) and move out of the basement.

brianhelba:
Merging this PR now, since it only changes TF, and that CI looks good.

Issue #200, opened 2020-10-30
Add endpoints for searching annotations in space


banesullivan:
Details to come -- this is for the ndsampler use case

------

Annotations are defined in pixel coordinates currently. We should add a way to search those annotations in world coordinates. Perhaps this could be done through a new `SpatialAnnotation` model that produces a convex whole of the features in a given `Annotation` and translates them accordingly based on the `RasterEntry`'s spatial reference for that `ImageEntry`

banesullivan:
We will need to store both pixel and geo-coordinates in the DB for the needs of another project

Issue #277, opened 2021-03-01
Annotations: endpoints for basic access and front-end viewer


banesullivan:
Realted to #200 but simpler

Add endpoints for accessing and creating annotations with an interface on the front-end for viewing

To do:
- [ ] Create serializers for the annotations with GET/POST endpoints
  - These should be accessed by image ID and annotation ID
  - List all annotations for an image: `api/geodata/imagery/image_entry/<int:image_pk>/annotations`
  - GET annotation by ID: `api/geodata/imagery/image_entry/<int:image_pk>/annotations/<int:ann_pk>`
  - POST new annotation: `api/geodata/imagery/image_entry/<int:image_pk>/annotations/new`
- [ ] Create annotation viewer functionality on front-end
  - Overlay annotations on raster detail view in map view
  - List all annotations on raster detail view
  - pop-up element to show annotation details (might wait until new VueJS front-ent)

PR #453, opened 2021-07-01
Add geospatial annotations

banesullivan:
Supporting work for https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/23

Related:
- Closes https://github.com/ResonantGeoData/ResonantGeoData/issues/200
- Addresses https://github.com/ResonantGeoData/ResonantGeoData/issues/277

Todo:

- We need to support ingesting [this annotation format](https://infrastructure.smartgitlab.com/docs/pages/api_documentation.html#region-model), which is essentially GeoJSON


banesullivan:
I'm considering moving this new `GeospatialTemporalAnnotation` model to the core app so that it can be used with any spatial data

PR #457, opened 2021-07-05
Overhaul ProcessedImage models and implement resampling support

banesullivan:
For https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/19

These changes add support for a few different use cases of `ProcessedImage` and the addition of a new function for resampling imagery by a numerical factor. The use cases we now support are:

1. Processing a single image (could be raster or non-raster) with a single output
2. Processing multiple images the same way (still 1-to-1 where each input corresponds to one output)
3. Processing multiple images with one or several image outputs (not a 1-to-1 mapping)

We need to support the case where downstream projects or users manually process the imagery and do not want/need a dedicated model to track the processing parameters but simply want to upload those new images linked to their source. To do so, I have made `ProcessedImage` non-abstract and combined our previous processing models into it with the addition of the  `ProcessedImageGroup` model for capturing the processing parameters.

Resolves https://github.com/ResonantGeoData/RD-WATCH/issues/4

### Other Changes

- Fixes to various serializers to make non-required fields actually non-required
- Make sure `related_name='+'` for various relationships to `ChecksumFile`
- Minor fix/adjustment to the status checking endpoint
- Spelling fixes 

### To Do

- [x] Update models for above use case/description
- [x] Update serializers, endpoints, admin actions, and other places where these models are used
- [x] Make sure new `ProcessedImageGroupSerializer` is being used by create view
- [x] Expose endpoints via Python client
- [x] Update processed image serializers in the fashion of #486 
- [x] Make sure Raster serializers (specifically STAC) link to the processed images as well


django-rgd-imagery/rgd_imagery/serializers/processed.py
```diff
@@ -1,59 +1,61 @@
...
+            'source_images',
         ]
 
     def create(self, validated_data):
-        """Prevent duplicated subsamples from being created."""
-        obj, created = models.RegionImage.objects.get_or_create(**validated_data)
-        if not created:
-            # Trigger save event to reprocess the subsampling
-            obj.save()
+        # TODO: why isn't this used by the creat view?
```
banesullivan, line 58: @mcovalt, I can't track down why the create view endpoint isn't using this serializer. Any help would be appreciated.

See `CreateProcessedImageGroup` in `django-rgd-imagery/rgd_imagery/rest/post.py`

django-rgd-imagery/rgd_imagery/serializers/__init__.py
```diff
@@ -2,19 +2,18 @@
 
 from .. import models
 from .base import ImageMetaSerializer, ImageSerializer, ImageSetSerializer
-from .processed import ConvertedImageSerializer, RegionImageSerializer
+from .processed import ProcessedImageSerializer
```
mcovalt, line 5:
```suggestion
from .processed import ProcessedImageSerializer, ProcessedImageGroupSerializer
```

django-rgd-imagery/rgd_imagery/serializers/__init__.py
```diff
@@ -2,19 +2,18 @@
...
 from .stac import STACRasterSerializer
 
 utility.make_serializers(globals(), models)
 
 __all__ = [
-    'ConvertedImageSerializer',
     'ImageSerializer',
     'ImageMetaSerializer',
     'ImageSetSerializer',
+    'ProcessedImageSerializer',
```
mcovalt, line 15:
```suggestion
    'ProcessedImageSerializer',
    'ProcessedImageGroupSerializer',
```

django-rgd-imagery/rgd_imagery/serializers/processed.py
mcovalt: Looks like the imports were missing. Would the changes I suggested resolve it?

banesullivan: That did it! Thanks! I would have thought that setting `serializer_class` in the view would be enough

banesullivan:
@mcovalt, would you please review the latest changes to the models in `django-rgd-imagery/rgd_imagery/models/processed.py`? I simplified it a bit further than we discussed this morning.

Now every `ProcessedImage` must belong to a `ProcessedImageGroup`. The `ProcessedImageGroup` defines the processing function `f` and `ProcessedImage` tracks the input(s) (`x_n`) and output (`y`). This  single implementation should work for all use cases

django-rgd-imagery/rgd_imagery/models/processed.py
```diff
@@ -1,94 +1,53 @@
...
-            RegionImage.SampleTypes.PIXEL_BOX,
-            RegionImage.SampleTypes.ANNOTATION,
-        ):
-            projection = 'pixels'
+    def _pre_delete(self, *args, **kwargs):
+        if self.processed_image:
+            self.processed_image.file.delete()
+        # TODO: clean up ancillary_files - this throws an error when done through the admin interface
+        for file in self.ancillary_files.all():
+            file.delete()
```
banesullivan, line 49: @mcovalt, can you help me figure out why this throws an error when deleting through the admin interface but not when calling `.delete()` on the object through the Python shell?

django-rgd-imagery/tests/test_rest.py
```diff
@@ -10,49 +10,78 @@ def test_download_image_file(admin_api_client, astro_image):
     assert status.is_redirect(response.status_code)
 
 
+@pytest.mark.xfail
```
banesullivan, line 13: I've marked two tests here as xfail. They shouldn't be failing and I cannot figure out what is different between these tests and what I do manually with swagger/the Python client.

I suspect something wacky is going on with the serializer where the m2m `source_images` field isn't saved when the serializer or view commits the record. @mcovalt, can you help me figure out what is different here?

django-rgd-imagery/rgd_imagery/models/processed.py
mcovalt: I'm not getting an error, but I'm also not getting any files deleted. What was your error?

mcovalt: What we really need is a FK w/ `CASCADE` from `ChecksumFile` to `ProcessedImage`. That would also enforce the constraint it seems we're assuming: one `ProcessedImage` per `ChecksumFile`.

mcovalt: I wonder if Django will manage the `CASCADE` for us if we put the FK on a `ChecksumFile` we multi-table inherit it in our app and put the FK there?

banesullivan: I'm very hesitant to start a multi-table inheritance for ChecksumFiles. We have relationships just like this one all over the place. Doing so would result in a lot of clutter, redundant tables, and just overall be hard to maintain. Let me double-check that this error is still occurring - I really think it shouldn't be and the current approach should suffice

banesullivan: I'm just going to not implement this for now and leave it on the user to clean up ancillary files (which is what we're doing for `Raster`'s ancillary files field too.

django-rgd-imagery/tests/test_rest.py
banesullivan: Deal with later

Issue #441, opened 2021-06-29
Improve and test PermissionPathMixin


banesullivan:
We should improve PermissionPathMixin so that we are not entirely hardcoding the paths everywhere.

We also need to test that the paths defined in the class are valid.

For https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/8

PR #463, opened 2021-07-07
Permissions path

venkatabhishek:
#441 

- Add testing for existence and correctness of `permissions_path` field (i.e on models that extend `PermissionPathMixin`): Iterates over models and traces path via class definitions
- Refactored `permissions_path` to not repeat `__collection__collection_permissions` everywhere: simply prepend to existing path. 

Point 2 could still be improved: One idea is to only explicitly have the next model in the path in the `permissions_path` field and dynamically build the rest. For example, the current path for class `FMVMeta` is [`fmv_file__file__collection__collection_permissions`], but instead we could change it to simply [`fmv_file` + <path for `class FMV`>] and then in class `FMV`, replace the path with [`file` + <path for `class File`>], etc.

### DIFF
--- a/django-rgd-3d/rgd_3d/models/point_cloud.py
+++ b/django-rgd-3d/rgd_3d/models/point_cloud.py
@@ -17,7 +17,7 @@ class PointCloud(TimeStampedModel, TaskEventMixin, PermissionPathMixin):
 
     data_link.allow_tags = True
 
-    permissions_paths = ['file__collection__collection_permissions']
+    permissions_paths = ['file']
 
 
 class PointCloudMeta(TimeStampedModel, PermissionPathMixin, DetailViewMixin):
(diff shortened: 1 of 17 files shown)


banesullivan (approved):
This is great work! Thanks, @venkatabhishek!!

> Point 2 could still be improved: One idea is to only explicitly have the next model in the path in the permissions_path field and dynamically build the rest. For example, the current path for class FMVMeta is [fmv_file__file__collection__collection_permissions], but instead we could change it to simply [fmv_file + <path for class FMV>] and then in class FMV, replace the path with [file + <path for class File>], etc.

This is an excellent idea! Would you want to implement this in a follow up PR, or do it here? I think it'd be nice to change `permissions_paths` to be a list of tuples in order to implement this. e.g.

```py
permissions_paths = [('file', ChecksumFile),]
```

where it gives the field name and the model to fetch the path from.

PR #465, opened 2021-07-07
time of day filter

### LINKED ISSUE #357: Time of day filter field
Add new acquisition date time of day filter.

Use case: find all rasters acquired between 8am and 4pm for all days in the acquired date range (i.e. only daytime imagery)

venkatabhishek:
Resolve #357 

- Added test in `test_filters.py` for time of day

### DIFF
--- a/django-rgd/rgd/filters.py
+++ b/django-rgd/rgd/filters.py
@@ -124,6 +124,11 @@ class SpatialEntryFilter(filters.FilterSet):
         return queryset
 
     def filter_time_of_day(self, queryset, name, value):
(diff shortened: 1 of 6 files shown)


banesullivan (approved):
I think this all looks good! Thanks, @venkatabhishek!

@mcovalt, would you be able to take a quick look?

banesullivan:
I added this to the search form on the front end and can confirm it is working as expected

banesullivan:
@venkatabhishek, can you add this to the Python client as well? It needs to be implemented in `search` and then have the same argument in the other search methods that wrap that

PR #467, opened 2021-07-08
Fix typos and add codespell to tox

banesullivan:
This fixes a handful of spelling errors and implements [codespell](https://github.com/codespell-project/codespell/) in a new tox test

I did not add this to the CI tests because codespell can get quite annoying if left on its own. This is useful as a local check to find and correct spelling errors.

simply run `tox -e codespell` to check for spelling errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 2D/3D/4D Geospatial Data API in Django
 
-A publically deployed instance of this application is available at https://www.resonantgeodata.com - find the project code for this at [ResonantGeoData/RD-OpenGeo](https://github.com/ResonantGeoData/RD-OpenGeo)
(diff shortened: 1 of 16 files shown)


banesullivan:
FYI @brianhelba

I'm not sure how useful codespell would be to other projects but it would be fairly easy/simple to include in the cookie-cutter. Let me know if you think this could be useful there

manthey:
I've found codespell to be useful, though it sometimes produces false positives (e.g., a variable called `IndexT` is mentioned in a comment and codespell complains that it should be `index`).

banesullivan:
It is exactly that reason why I did not enable it on the CI - would lead to tons of failed checks. I've found that you typically have to build a list of ignored words per project to avoid the false positives which can be a burden on developers.

PR #469, opened 2021-07-12
fix inline geo admin widget

### LINKED ISSUE #391: GeoAdmin widget incorrectly rendering geometry
The `OpenLayersWidget` we are using in the admin view of the geo Django models is incorrectly rendering geometry. I suspect this is a simple coordinate issue where we need to transform the geometry before passing it to the model admin widget:

https://user-images.githubusercontent.com/22067021/117475485-5761c300-af19-11eb-97b5-7cd8e29bf14c.mov

venkatabhishek:
Resolve #391

### DIFF
--- a/django-rgd-imagery/rgd_imagery/admin/base.py
+++ b/django-rgd-imagery/rgd_imagery/admin/base.py
@@ -2,6 +2,8 @@ from django.contrib import admin
 
 # from django.contrib.admin import SimpleListFilter
 from django.contrib.gis.admin import OSMGeoAdmin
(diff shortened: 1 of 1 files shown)


banesullivan:
This appears to have changed the basemap of the widget but the coordinate system is still off. The example datasets appear in the wrong place (near 0º, 0º)

Screen Shot 2021-07-13 at 9 08 18 AM

banesullivan (approved):
Woohoo! The latest changes fixed this! Thanks, @venkatabhishek 

Screen Shot 2021-07-21 at 9 49 18 AM

PR #470, opened 2021-07-12
Client testing

### LINKED ISSUE #443: Proper testing for rgd_client
We need to implement real tests for the Python client

The tests for this would require running RGD with data prepopulated in the background then executing the client tests

venkatabhishek:
Resolve #443 

- Add new tox environment for testing client
- `conftest.py`: Resets db, runs migrations, and starts dev server in background
- Add fixtures for running populate database (i.e. management commands such as 'rgd_3d_demo')
- Added tests for basic usage (i.e. as show in README)
- Add ci test flow for client


rgd-client/tests/data_fixtures.py
```diff
@@ -0,0 +1,28 @@
...
+
+from . import MANAGE_PATH, PYTHON_PATH
+
+sources = [
+    'rgd_3d_demo',
+    'rgd_fmv_demo',
+    'rgd_fmv_wasabi',
+    'rgd_geometry_demo',
+    'rgd_imagery_demo',
+    'rgd_imagery_landsat_rgb_s3',
```
banesullivan, line 13: Let's definitely not include `rgd_imagery_landsat_rgb_s3` or `rgd_fmv_wasabi` when testing as these are larger demo datasets not meant for tests

These two are likely the reason it is running so slow

banesullivan (commented):
So far this is looking great!

rgd-client/rgd_client/utils.py
```diff
@@ -49,7 +49,7 @@ def limit_offset_pager(session: Session, url: str, **kwargs) -> Generator[Dict,
         r.raise_for_status()
 
         # Yield results
-        results = r.json()['results']
+        results = r.json()
```
banesullivan, line 52: @AlmightyYakob, I've had to change this back and forth a few times. How should this be?

AlmightyYakob: It should be `results = r.json()`, otherwise you're just yielding a dict, and need to manually extract the `results` field. @venkatabhishek What was the reason for this change?

venkatabhishek: When I print the contents of `r.json()` when running something like `client.search(query=json.dumps(bbox), predicate='intersects')` as in the README example, it is just a list of the returned spatial entries. With `r.json()['results']` it raises an error related to this (i.e. TypeError: list indices must be integers or slices, not str)

AlmightyYakob: The search endpoint returns a paginated response, not a list directly, correct @banesullivan? What you're observing @venkatabhishek implies that the search endpoint is returning a list directly.

banesullivan: Hmm.. I'm not exactly sure what is happening here. The search endpoints use `rest_framework.generics.ListAPIView` which I am observing returning a list directly. It seems to me that using `limit` in the Python client to paginate the search results isn't currently doing anything?

AlmightyYakob: Okay my bad, it seems the production is not up to date with that endpoint change.

The `limit_offset_pager` function is intended to be used on endpoints which are making use of limit offset paging. If the search endpoint isn't intended to paginate the results, then we shouldn't use this function, and should just call that endpoint directly in `Rgdc.search`.

I think the easiest thing to do right now is to change this line back to `results = r.json()['results']`, and remove the call to `limit_offset_pager` in `Rgdc.search` (and anywhere else it may be necessary). If pagination is meant to be applied to that endpoint and fixed later on, then these tests should fail (:tada: yay tests) and we can add the use of `limit_offset_pager` back in. Regardless, the api surface for this method shouldn't change (it still returns a list in either case).

Does this sound good to both of you @banesullivan @venkatabhishek?

AlmightyYakob: Addressed in bdb6b8c so I'm happy

banesullivan (approved):
@venkatabhishek, great work here! It is looking excellent and I'll get this merged in soon here

Issue #480, opened 2021-08-10
GRB data ingest


banesullivan:
See email.

I have some GRB climate data and a script to plot those. The task is to index the GRB data in RGD and see if we can add a web viewer (perhaps using VTK.js first)

banesullivan:
As mentioned in #482, GRIB is supported by GDAL and we can serve tiles of all the bands in these datasets on a map view. Unfortunately, large_image isn't extracting the right spatial reference and thus the tile serving isn't working.

I am working to add some new models to the 3D module for loading a subset of the data in these GRIB files into a 3D volume (`vtkImageData`) that will be rendered in 3D with VTK.js

banesullivan:
https://github.com/girder/large_image/pull/639 solved an issue that was preventing tile serving for these files, though there are still some kinks to work out.

These data can be generally previewed on RGD's front-end if loaded as a raster after #482

----


I have some other work in progress to render these as 3D volumes using VTK.js. I need to update our VTK.js viewer to support ImageData but it will look something like this.

<img width="1807" alt="Screen Shot 2021-08-16 at 3 25 57 PM" src="https://user-images.githubusercontent.com/22067021/129631546-2b06b089-073f-40b5-80ad-7f29047fad5d.png">


This uses some hacky code to stack all 2D layers of each data variable in the dataset into a 3D volume.

Issue #421, opened 2021-06-03
RGB Raster View


banesullivan:
We should add a new raster detail view page for previewing the raster in a RGB color mode. I'm thinking the place to start is by specifying this as a new URL:

```
geodata/raster/<int:pk>/rgb/<str:r_params>/<str:g_params>/<str:b_params>
```

This URL would allow the user to specify exactly which ImageEntry's to use from the raster and then which bands within each of those image entries/files to use.

a few different use cases:

- RGB bands all coming from one file (one ImageEntry) in the ImageSet
- RGB bands coming from one band in 3 different files (ImageEntries in the ImageSet)
- then a mix of these two (2 bands from one file and 1 band from another file)

Potential challenge: we do not enforce that a raster's images all be the same resolution or even coverage. Most often they are, but we need to be able to catch the case when they are not and throw an error.

for each of the `str` band params, we could pass the image entry ID `<image_entry_id>` which defaults to the first band in that file, then to specify the band within that file, `<image_entry_id>-<band_index>`

This is just a rough idea of how to do this.

@manthey, can large image support taking three different files and combing bands in them to generate RGB tiles?

aashish24:
Is this for SMART or IH? In general this is a nice feature to have

manthey:
Not currently.  It should be a straightforward addition to make large_image do this.  Currently, you composite bands via something like `style={"bands": [{"band": 1, "palette": ["black", "red"]}, {"band": 2, "palette": ["black", "green"]}]}`.  This would need something like `style={"bands": [{"band": 1, "palette": ["black", "red"]}, {"source": "<some filename>", "band": 1, "palette": ["black", "green"]}]}`.

The other way to do this is with gdal vrt files and use the vrt file as the source.

Go ahead and add a issue to large_image.

banesullivan:
> Is this for SMART or IH? In general this is a nice feature to have

No project specifically -- I believe SMART would benefit from it. I mainly had the idea and wanted to put it on the feature request backlog for later

banesullivan:
This would require upstream features in large_image (ref https://github.com/girder/large_image/issues/680) and some rethinking of our tile endpoints. Since no one is pushing this, I am closing for now

manthey:
For the record, this could be accomplished currently using a vrt file to combine two tile sources.  Otherwise, there is another project which *might* end up allow tiles from multiple files to be composited together.

PR #482, opened 2021-08-16
Support individual band selection in raster detail view

banesullivan:
In support of https://github.com/ResonantGeoData/ResonantGeoData/issues/480 (GRIB data files are supported by GDAL, so we can technically treat these as rasters and preview individual bands). At present, our ETL routines are not properly extracting the spatial reference of GRIB datasets, so this doesn't work. GRIB support will come in its own PR.

This is a needed feature on its own so I'd like to land this on its own and sort of moves us toward https://github.com/ResonantGeoData/ResonantGeoData/issues/421


banesullivan:
The template-side (HTML/JS) of this is kind of dirty, but we'll soon completely overhaul the front-end, so I'm not going to spend time optimizing it

PR #483, opened 2021-08-17
Add new GRIB model with ETL routine to create 3D volume

banesullivan:
Leaving this as a draft PR as I don't think this feature is something we should ship.

We should use `large_image` for serving tiles of these GRIB datasets on a 2D geospatial map


banesullivan:
Closing this in favor of #484

PR #486, opened 2021-08-18
More POST (and GET) endpoints!

### LINKED ISSUE #452: Improve/add POST endpoints for better data ingest
We need to add a complete set of POST endpoints for data ingestion/upload

Models to either create a serializer for or to update the existing serializer to handle POST requests:

- [x] Collection and CollectionPermission
- [x] ChecksumFile
- [x] SpatialAsset
- [x] Image
- [x] ImageSet
- [x] ImageSetSpatial
- [x] Raster
~- [ ] Annotation / Segmentation / PolygonSegmentation / RLESegmentation~
- [x] GeometryArchive
- [x] Geometry
- [x] FMV
- [x] PointCloud
- [x] PointCloudSpatial

Once these are implemented, we will need to follow up with adding a method for each endpoint in the Python client

banesullivan:
Partially address #452 

This adds the POST endpoints we hope to expose in the Python client for the work in https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/29

Adds new POST endpoints for (and GET when not already present):

- `Collection` and `CollectionPermission`
- `ChecksumFile` (URL files only)
- `SpatialAsset`
- `Image`
- `ImageSet`
- `ImageSetSpatial`
- `Raster`

cc @AlmightyYakob

### DIFF
--- a/django-rgd-imagery/rgd_imagery/rest/get.py
+++ b/django-rgd-imagery/rgd_imagery/rest/get.py
@@ -29,6 +29,12 @@ class GetImageSet(RetrieveAPIView, _PermissionMixin):
     queryset = models.ImageSet.objects.all()
 
 
+class GetImageSetSpatial(RetrieveAPIView, _PermissionMixin):
+    serializer_class = serializers.ImageSetSpatialSerializer
+    lookup_field = 'pk'
+    queryset = models.ImageSetSpatial.objects.all()
+
+
 class GetRasterMeta(RetrieveAPIView, _PermissionMixin):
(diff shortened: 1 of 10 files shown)


mcovalt (approved):
I've tested all these out and they work well!

In the future we may want to clean things up with [viewsets](https://www.django-rest-framework.org/api-guide/viewsets/). We currently have REST endpoints separated logically via REST verb, but that would group all those together under a model.

banesullivan:
Ah, viewsets are a great idea! I'll open an issue under tech-debt about this

PR #492, opened 2021-08-20
Add new created_by field to ChecksumFile

### LINKED ISSUE #474: Default collection for user uploads
We have users that need to be able to upload data to different collections through the Python client.

First, we should set a default collection for each user so that data uploaded by them belong to that collection.

Then, we need to expose functionality in the endpoints/python client that enables users to set the collection for each upload.

banesullivan:
Resolve #474 

Todo

- [x] handle permissions filtering

After merging, we will need to make sure all downstream projects have `django-crum` installed as middleware


banesullivan:
@mcovalt, please review - I have added a way to handle permissions filtering for the case of files being `created_by` a user but not necessarily being in a Collection for which that user has access

mcovalt (approved):
These changes look good to me and work as expected. Thanks for handling the case where `created_by` is only set when initially saving, not on subsequent saves.

PR #493, opened 2021-08-20
Fix error in order_datetimes, add doctest

Erotemic:
The previous usage of `order_datetimes` used the syntax `value1, value2 if value1 < value2 else value2, value1`, which actually will return a 3-tuple of (value1, max(value1, value2), value2) because the comma operator has lower precidence than inline if staements.

I also added a doctest with the code I used to verify that it was failing and then ensure that the fix worked. 

We should be able to get this test to run in the CI without too much trouble: Add `xdoctest` as a dependency and give `--xdoctest` as an argument to pytest. I haven't used tox in awhile, so not sure if that complicates anything.

### DIFF
--- a/rgd-client/rgd_client/utils.py
+++ b/rgd-client/rgd_client/utils.py
@@ -94,9 +94,18 @@ def datetime_to_time(value: object):
 
 
 def order_datetimes(value1: object, value2: object):
-    """Sort 2 objects if they are datetimes."""
+    """
+    Sort 2 objects if they are datetimes.
+
+    Example:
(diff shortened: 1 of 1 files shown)


banesullivan:
Thanks, @Erotemic! 

re xdoctest: We should just be able to add `--xdoctest` to the `addopts` section of `[pytest]` at the end of the tox file as well as adding as a dependency in `requirements.txt`

Erotemic:
Made a new PR in https://github.com/ResonantGeoData/ResonantGeoData/pull/494 to see how easy it is to add xdoctest to the test suite.

PR #494, opened 2021-08-20
Dev/xdoctest

Erotemic:
Builds on #493 

Adds declaration of xdoctest as a test dependency and adds the flags to tox. I also added a second simple doctest  to `spatial_search_params`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,6 @@ pytest-factoryboy
 pytest-mock
 pytest-cov
 pytest-memprof
+xdoctest
 -e ./django-rgd
 -e ./testing-utils
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ pytest-factoryboy
 pytest-mock
 pytest-cov
 pytest-memprof
-xdoctest
+xdoctest >= 0.15.6
 -e ./django-rgd
 -e ./testing-utils
--- a/rgd-client/rgd_client/utils.py
+++ b/rgd-client/rgd_client/utils.py
@@ -168,6 +168,29 @@ def spatial_search_params(
     offset: Optional[int] = None,
     time_of_day: Optional[DATETIME_OR_STR_TUPLE] = None,
 ) -> Dict:
+    """
+    Validate, fixup, and package parameters for an RGD search
+
+    Example:
+        >>> query = {
+        >>>     "type": "Polygon",
+        >>>     "coordinates": [
+        >>>         [
+        >>>             [-81.7745, 33.1346],
+        >>>             [-81.7746, 33.1437],
+        >>>             [-81.7744, 33.1437],
+        >>>             [-81.7666, 33.1436],
+        >>>             [-81.7745, 33.1346]
+        >>>         ]
+        >>>     ]
+        >>> }
+        >>> predicate = 'intersects'
+        >>> acquired = (
+        >>>     datetime.fromisoformat('2101-01-01T00:10:00'),
+        >>>     datetime.fromisoformat('2101-01-01T00:23:00')
+        >>> )
+        >>> params = spatial_search_params(query, predicate, acquired=acquired)
+    """
     # The dict that will be used to store params.
     # Initialize with queries that won't be additionally processed.
     params = {
--- a/rgd-client/rgd_client/utils.py
+++ b/rgd-client/rgd_client/utils.py
@@ -94,9 +94,18 @@ def datetime_to_time(value: object):
 
 
 def order_datetimes(value1: object, value2: object):
-    """Sort 2 objects if they are datetimes."""
+    """
+    Sort 2 objects if they are datetimes.
+
+    Example:
+        >>> from rgd_client.utils import *  # NOQA
+        >>> value1 = datetime.fromisoformat('2000-01-01T00:13:30')
+        >>> value2 = datetime.fromisoformat('1999-01-01T00:13:30')
+        >>> result = order_datetimes(value1, value2)
+        >>> assert len(result) == 2
+    """
     if isinstance(value1, datetime) and isinstance(value2, datetime):
-        return value1, value2 if value1 < value2 else value2, value1
+        return (value1, value2) if value1 < value2 else (value2, value1)
 
     return value1, value2
 
--- a/rgd-client/rgd_client/utils.py
+++ b/rgd-client/rgd_client/utils.py
@@ -169,7 +169,7 @@ def spatial_search_params(
     time_of_day: Optional[DATETIME_OR_STR_TUPLE] = None,
 ) -> Dict:
     """
-    Validate, fixup, and package parameters for an RGD search
+    Validate, fixup, and package parameters for an RGD search.
 
     Example:
         >>> query = {
--- a/rgd-client/setup.py
+++ b/rgd-client/setup.py
@@ -45,5 +45,5 @@ setup(
         'geomet',
         'tqdm',
     ],
-    extras_require={'dev': ['ipython']},
+    extras_require={'dev': ['ipython', 'xdoctest']},
 )
--- a/rgd-client/setup.py
+++ b/rgd-client/setup.py
@@ -45,5 +45,5 @@ setup(
         'geomet',
         'tqdm',
     ],
-    extras_require={'dev': ['ipython', 'xdoctest']},
+    extras_require={'dev': ['ipython']},
 )
--- a/tox.ini
+++ b/tox.ini
@@ -132,7 +132,7 @@ deps =
     -e ./django-rgd-imagery[worker]
     -e ./rgd-client[dev]
 commands =
-    pytest --xdoctest {posargs}
+    pytest {posargs}
     # coverage html -d {toxworkdir}/htmlcov/django-rgd-client
     # coverage xml -o {toxworkdir}/coverage.xml
 
@@ -171,7 +171,7 @@ ignore =
 
 [pytest]
 DJANGO_SETTINGS_MODULE = test_project.settings
-addopts = --strict-markers --showlocals --verbose --durations=0
+addopts = --strict-markers --showlocals --verbose --durations=0 --xdoctest 
 filterwarnings =
     ignore:.*default_app_config*.:django.utils.deprecation.RemovedInDjango41Warning
     ignore::DeprecationWarning:minio
--- a/tox.ini
+++ b/tox.ini
@@ -132,7 +132,7 @@ deps =
     -e ./django-rgd-imagery[worker]
     -e ./rgd-client[dev]
 commands =
-    pytest {posargs}
+    pytest --xdoctest {posargs}
     # coverage html -d {toxworkdir}/htmlcov/django-rgd-client
     # coverage xml -o {toxworkdir}/coverage.xml
 


Erotemic:
I've got xdoctest installed and running alongside pytests, but currently it is unable to find the doctests. The reason is because the tox entry for `[testenv:test-rgd-client]` does a changedir to `changedir = rgd-client/tests`, so when pytest does its discovery for tests it only looks in that CWD and never sees the source directory. For xdoctest to discover the doctests the source for the module needs to either be in the CWD pytest is run in, or passed to pytest.

Often I just run pytest in the root of the my repo, so this isn't a problem, but there are instances where I'm testing wheels where it is more desireable to run pytest in a sandboxed directory (e.g. I do this in kwimage).

I have a semi-hacky way of handling this case at the moment: https://gitlab.kitware.com/computer-vision/kwimage/-/blob/master/.gitlab-ci.yml#L183 At a high level, the CI has already built the wheel and pip installed the wheel, so I don't want to run pytest in the root, because I want it to pick up the contents of the installed wheel, not the source. So I cd into an empty sandbox directory, and run pytest and point at `../tests` (because the test dir is also in the root). Then to give it the path to the installed source I want to test (because it needs to parse the doctests from that). I use `$(python -c "import ubelt; print(ubelt.modname_to_modpath('kwimage'))")`, which returns the path to where the wheel is installed.

I'm not sure how this is best handled in this case. If the commands in the "commands" section of the tox file are just run in a bash shell, then the path to the installed version of "rgd_client" could be obtaine via: 

```
$(python -c "import rgd_client, os; print(os.path.dirname(rgd_client.__file__))")
```

which has the side effect of importing the module we want to test, or which is able to obtain the path statically without importing the module 

```
$(python -c "import ubelt; print(ubelt.modname_to_modpath('rgd_client'))")
```

which works via pathlib and is roughly equivalent to this code: 

```
$(python -c "import pkgutil, os; print(os.path.dirname(pkgutil.find_loader('rgd_client').get_filename()))")
```

It also might be fine to pass it "../rgd_client", but I believe xdoctest will attempt to import that module from that location in order to test it (I would have to double check on that to be sure), so that might cause issues --- especially with coverage --- if we are really trying to test the installed version.

banesullivan (commented):
Hm. With some fresh eyes on this, I'm thinking that we should *not* have xdoctest in all test suites. It does not make sense for the other five Django Python packages but only makes sense for the Python client package in special cases (xdoctest would have to handle setting up the Django server, database, etc.). Thus, we should just install it as a dependency in the client test env section and run xdoctest in the fashion of: `python -m xdoctest /path/to/your/pkg-or-module.py` to avoid the directory issue.

I've gone ahead and made these changes and pushed them here.

However, I am still hesitant to add xdoctest. It can only be used with the internal helper methods in the client package (which are not a part of the user-facing API and thus do not need fancy docstrings). I would like to avoid the inconsistency of having some tests in pytest modules and some tests in docstrings. It may get confusing for new devs on the project when they see these doctests and start adding them to other parts of the client code which rely on a Django server (these would fail). Further the point of doctstrings like these are to make sure the API itself is tested in a way in which the user might use it and since these are internal methods, they are highly subject to change.

Personally, I think xdoctest adds too much overhead here and I'm in favor of using plain old pytests. @ResonantGeoData/maintainers, thoughts?

Otherwise, xdoctest looks like an awesome expansion on doctest and I'm thinking I should switch some of my other projects already using doctest to it!

banesullivan:
Just realized my changes weren't added here. Heres the diff:

```
diff --git a/requirements.txt b/requirements.txt
index c52e36d..ad965c3 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,5 @@ pytest-factoryboy
 pytest-mock
 pytest-cov
 pytest-memprof
-xdoctest >= 0.15.6
 -e ./django-rgd
 -e ./testing-utils
diff --git a/tox.ini b/tox.ini
index e2c8df9..4fbea6d 100644
--- a/tox.ini
+++ b/tox.ini
@@ -125,6 +125,7 @@ commands =
 changedir = rgd-client/tests
 deps =
     -r requirements.txt
+    xdoctest>=0.15.6
     -e ./django-rgd
     -e ./django-rgd-3d[worker]
     -e ./django-rgd-fmv[worker]
@@ -133,6 +134,7 @@ deps =
     -e ./rgd-client[dev]
 commands =
     pytest {posargs}
+    python -m xdoctest ../rgd_client
     # coverage html -d {toxworkdir}/htmlcov/django-rgd-client
     # coverage xml -o {toxworkdir}/coverage.xml

@@ -171,7 +173,7 @@ ignore =

 [pytest]
 DJANGO_SETTINGS_MODULE = test_project.settings
-addopts = --strict-markers --showlocals --verbose --durations=0 --xdoctest
+addopts = --strict-markers --showlocals --verbose --durations=0
 filterwarnings =
     ignore:.*default_app_config*.:django.utils.deprecation.RemovedInDjango41Warning
     ignore::DeprecationWarning:minio
```

Erotemic:
It very well could be the case that it does not make sense here. Hence why I put it in this PR instead of the previous one.

> the point of doctstrings like these are to make sure the API itself is tested in a way in which the user might use it

That is _a_ point of docstrings / doctests, another is reducing the overhead of writing tests, so devs actually write them. Pytest with all the fixtures and magic can be a real pain. Personally, just switching between files (let alone files in different directories) is enough overhead to inhibit me from writing a test, especially a small one like the one that could have prevented the error in #493.  Coupling the test with the code means that its easier to change the tests for pieces that are subject to change.

That being said, not everything is suitable for a doctest, but its usually eaiser to migrate an existing doctest to a separate unit test than it is to deal with projects that don't have tests. My main philosophy on the subject is to lower the barrier to entry for tests as much as possible.

On the topic of fixtures: Where does `live_server` come from? in rgd-client/tests/conftest.py? I greped for it but found no other instance. If there was a way to make a "demo" or "dummy" `RgdcSession`, then doctests would start making a more sense.

banesullivan:
> another is reducing the overhead of writing tests, so devs actually write them. 

Fair point... considering the fact that we have all of 3 tests, we could use as much help as we can get writing more.

I'll have to continue giving this some thought so that we do this in a way that makes the most sense for this project. Though I still feel strongly that we shouldn't fragment our tests across pytest modules and docstrings in this case.

> Where does live_server come from?

`live_server` is a fixture from pytest-django: https://pytest-django.readthedocs.io/en/latest/helpers.html#live-server

Spinning this up more than once (once for our pytest modules and once for xdoctest) would be too expensive (I'm assuming this would have to happen a second time in a new context for xdoctest)

Erotemic:
Ugg, fixtures are such a rabbit hole. Is it so wrong to want to just open up IPython and have the ability to create an instance of one? 

I looked up the definition for "live_server", which has it's own argument `request`, which has absolutely no way to easilly see where that is coming from or what it's attributes might be.

If I haven't mentioned, I really don't like fixtures.

I'm not advocating for this approach, more just putting it out there as a dicussion point, but if I was testing this, my inclination would be to have a "demo.py" file that could prop up a demo "live_server". The function would be memoized so it wouldn't be executed more than once, and then instead of using magic fixtures, I would call the function to get the instance. That way a new dev could trace back where everything came from. The `atexit` module could be used for handling shutdown logic. This doesn't quite work, but this is the idea: 

```python
import ubelt as ub

@ub.memoize
def demo_live_server():
    from pytest_django import live_server_helper
    import os
    # Source code contains no indication of what request is, ignoring it
    # request.config.getvalue("liveserver")

    # HACKS:
    # probably need to setup some background database?
    # information in rgd_testing_utils/settings.py seems important
    import sys
    sys.path.append('/home/joncrall/code/ResonantGeoData/testing-utils/')
    sys.path.append('/home/joncrall/code/ResonantGeoData/rgd-client/tests/')
    os.environ['DJANGO_SETTINGS_MODULE'] = 'test_project.settings'
    # from django.core.management import execute_from_command_line
    # Need to do some sort of execute_from_command_line?

    addr = os.getenv(
        "DJANGO_LIVE_TEST_SERVER_ADDRESS"
    ) or "localhost"

    server = live_server_helper.LiveServer(addr)
    import atexit
    atexit.register(server.stop)
```

Of course it's possible that I'm not considering all the complexities of network programming, and in that case fixtures are really great once you get used to them. Still, I wish they were easier to trace.

brianhelba:
@Erotemic pytest is a pretty popular test framework for Python (7.7k stars on GitHub). There is some magic to be learned, and I'd suggest reviewing some of the basic docs ( https://docs.pytest.org/en/latest/explanation/fixtures.html ) and trying to contribute some RGD functionality using the existing test infrastructure (using `pytest-django`, etc.). In particular, note that running `pytest --fixtures` lists all of the available fixtures for a given test environment.

If it truly looks like pytest is not a good fit for RGD, then perhaps it's worth opening a separate issue to discuss. However, I'm concerned that this PR is the start of a schism, where some developers will write tests with pytest and others write them with xdoctest. That seems unhealthy for RGD as a collaborative effort.

AlmightyYakob:
I don't like the idea of splitting our tests across both `pytest` and `xdoctest`. It's somewhat confusing to have them split up, and how/where to write tests could become ambiguous. I don't see a compelling reason not to just use `pytest` for any unit tests. It's a little more effort to write pytest tests, but it allows for much more comprehensive tests, sometimes without much boiler plate code thanks to fixtures, parametrization, etc. At the end of the day, it's our responsibility to ensure that tests are added alongside any code that we think needs them, and I think pytest provides a reasonably low barrier to entry.

zachmullen:
I think doctest makes sense _if you were wanting to write example code in your docstrings anyway_. It should not be viewed as an alternative unit testing library to pytest, but rather as something that validates correctness of docstrings.

Judging by the "Documentation" section of the current README in this repository, it seems like the documentation plan for RGD isn't even fleshed out yet. If that's the case, I think this addition is premature.

Erotemic:
@brianhelba I'm familiar with pytest. XDoctest itself is a pytest plugin, and is the main way in which it is used. I didn't mean to suggest that pytest isn't a good fit. Pytest is the de-facto Python tester for good reason. My only gripe with it is that fixtures are too opaque and magical. Specifically: (1) using the name of an argument to define how things are connected is far too implicit _for my taste_ (2) there is no good way to instantiate one in a standalone way, it requires that you are already running a pytest session, which is really my main problem with it.

I don't think there is any risk of a schism. I'm more than happy to let this PR die, but I do think the discussion is intersting, and I'd like to continue it. 

@AlmightyYakob 

>  I don't see a compelling reason not to just use pytest for any unit tests. It's a little more effort to write pytest tests

Yet, developer's don't make that effort. Consider the example of [order_datetimes](https://github.com/ResonantGeoData/ResonantGeoData/pull/493/files). It is a non-public facing function that 
may get refactored away. If a small test is _coupled_ with the function, you detect or prevent bugs like the one in #493 and if you refactor it, you don't have to worry about finding and deleting the associated unit test somewhere else. It's generally code you would write anyway as you are developing the function, so as long as it's not too cumbersome, I don't see any reason to not just stick it in the docstring as a doctest and get a bit of extra coverage "for free".

That being said, you could also just stick it in a unit test as well, but I do think having _some tests coupled with the function is a more powerful idea than people give it credit for. It makes it much easier to move functions between libraries/projects and ensure they are tested.

Again, it's not a replacement for pytest, but it is supplemental.

> It's somewhat confusing to have them split up, and how/where to write tests could become ambiguous. 

As long as you are executing both, it doesn't really matter at the end of the day, and it becomes a question of style and opinion. My opinion is that you should try to show a developer how to use your function in the function itself, and that should be done by giving them code that is easy to copy/paste into IPython and have it execute. Once they have executing code, they can play with it. 

When it comes to writing robust tests cases with parameterization (a fixture that doesn't bother me because it's clear how things are connected in the code itself), then separate unit tests and system tests should live in a sepaarate file, if only for the reason that they would bring too much clutter into the code itself.

For instance in ubelt, every public function has a doctest, but that is not enough to achieve 100% coverage. There is an additional "tests" folder containing unit tests for pytest. While I can be somewhat lax with just how [much clutter I'll allow in a doctest](https://github.com/Erotemic/ubelt/blob/master/ubelt/util_cmd.py#L197), the full comprehensive test suite does belong in a [seperate unit test file](https://github.com/Erotemic/ubelt/blob/master/tests/test_cmd.py). Often times, I'll start out writing the unit test in a doctest, but if it grows to large I can move it out of the file. This pattern lowers the barrier to writing the most minimal test, and once you have a minimal test, it's a lot easier to write the comprehensive one. (But even if you don't you at least you know there is some path through your code that doesn't error).

@zachmullen 

> It should not be viewed as an alternative unit testing library to pytest, 

Yes. You generally want to make unit tests more robust than something you would put in a doctest.

> but rather as something that validates correctness of docstrings

I disagree there. Doctests sever as an example that demonstrates how to execute the code. One of the things I like most about them is that they provide an entrypoint into the code. They define all of the inputs needed to call a function, so among other things - that allows a developer to drop into IPython, copy over the demo data, and start playing with the internal logic of the function.  

Again, in the context of this project requiring that a server be available to interact with, it might not make sense to execute doctests in a CI Suite, but if you had some `rgd_client.demo` module that could stand up a backend test server, you could add something like this to the public facing API

```python
class Rgdc:
    """
    Example:
        >>> from rgd_client
        >>> api_url = rgd_client.demo.ensure_demo_server()
        >>> self = Rgdc(api_url)
        >>> query = {
        >>>     "type": "Polygon",
        >>>     "coordinates": [
        >>>         [
        >>>             [-81.7745, 33.1346],
        >>>             [-81.7746, 33.1437],
        >>>             [-81.7744, 33.1437],
        >>>             [-81.7666, 33.1436],
        >>>             [-81.7745, 33.1346]
        >>>         ]
        >>>     ]
        >>> }
        >>> predicate = 'intersects'
        >>> acquired = (
        >>>     datetime.fromisoformat('2101-01-01T00:10:00'),
        >>>     datetime.fromisoformat('2101-01-01T00:23:00')
        >>> )
        >>> results = self.search(query=query, predicate=predicate, acquired=acquired)
        >>> for result in results:
        >>>     print('result = {!r}'.format(result))
    """ 
```

So then the API docs would provide users with an _executable_ example that demonstrates how to use the code. And if you have these, you should execute them in your test suite (as you said to validate the correctness of the example in your docstring). But even though it is not a robust unit test, just blindly executing a code path and not seeing (or seeing) a crash is useful  diagnostic information that I would argue you should consider as part of your test suite.

(I also wonder if hesitence towards doctests has to do with how github and some editors don't higlight them: https://github.com/rouge-ruby/rouge/issues/920)

aashish24:
is this PR stalled @Erotemic @banesullivan

banesullivan:
It is my opinion that the introduction of doctest/xdoctest adds too much complexity and fragmentation to our tests. Additionally, because we are working with Django and the burden of spinning up the app/web server during testing, it would be too much of an obligation to maintain a separate testing infrastructure for xdoctest when we have a great solution from pytest-django. 

Further, it is not (yet) our goal to provide meaningful docstrings as most of what we are providing here is a web API and not a Python API intended for direct use (IMO, doctests in our Django models/utils are frivolous). We are still working out what kinds of documentation we will offer, and unfortunately, the timing isn't good for this. With that said, there could be a place for this specifically for the Python client down the road (where docstrings do make sense).

For these reasons, I am closing the PR until further notice. Thank you, @Erotemic, for provoking this discussion. There's still a lot left unaddressed here; hopefully, we can take on some of these challenges later on when the app is more stable, and we have a clear roadmap for our documentation/testing.

PR #496, opened 2021-08-23
Pin pooch to <1.5.0

AlmightyYakob:
Due to a change in [`pooch`](https://github.com/fatiando/pooch/issues/252), `scikit-image` is [broken](https://github.com/scikit-image/scikit-image/issues/5528) on import. This was [fixed](https://github.com/scikit-image/scikit-image/pull/5529) in `scikit-image`, but a release hasn't been cut yet. This currently breaks our tests.

The breaking change was introduced in version `1.5.0` of `pooch`, so PR restricts the version accordingly.

### DIFF
--- a/django-rgd/setup.py
+++ b/django-rgd/setup.py
@@ -61,7 +61,8 @@ setup(
         'djangorestframework',
         'drf-yasg',
         'GDAL',
-        'pooch',
+        # TODO: Unpin once scikit-image and kwimage are updated
+        'pooch<1.5.0',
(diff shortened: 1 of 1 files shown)


banesullivan (approved):
I figured scikit-image would cut a release today but looks not. Merging...

manthey:
Pooch just released 1.5.1 to unbreak scikit-image.  scikit-image has a fix, but I don't know when they'll publish a release (probably not for a while now that Pooch has done so).

PR #504, opened 2021-08-26
Redesign Python Client

AlmightyYakob:
Closes https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/20

This majorly redesigns the python client in the following ways:

* Modularize the client, allowing for plugin clients (both internal and external) to be registered via the `rgd_client.plugin` entry point
* Closely couple client plugins with their respective django rgd modules, allowing for the sharing of test suites and fixtures

TODO:
- [x] Update READMEs / add documentation to specify the plugin registration and use steps


mcovalt (approved):
I spent quite a bit of time testing this out. I think the `rgd_client.plugin` entry point is awesome. The `py_client` fixture being declared and tested next to the rest of the tests makes the client much more integrated with the rest of the code.

Overall, a lot of great design decisions. I would like the `py_client` declared once in the `django-rgd` module as you've noted with a TODO, but I think  for now it's OK to repeat. Can you envision a case where this fixture would differ between modules in RGD?

.github/workflows/release.yml
```diff
@@ -10,10 +10,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rgd-module: ['django-rgd', 'django-rgd-3d', 'django-rgd-fmv', 'django-rgd-geometry', 'django-rgd-imagery', 'rgd-client']
+        rgd-module: ['django-rgd', 'django-rgd-client', 'django-rgd-3d', 'django-rgd-fmv', 'django-rgd-geometry', 'django-rgd-imagery']
         include:
           - rgd-module: django-rgd
             TWINE_PASSWORD_VAR: PYPI_API_TOKEN_RGD
+          - rgd-module: django-rgd-client
+            TWINE_PASSWORD_VAR: PYPI_API_TOKEN_RGD_CLIENT
```
banesullivan, line 18: I think here, you want to do `django-rgd/client` instead of `django-rgd-client` to set the right working directory.

banesullivan: I think we also need to publish the `rgd-imagery-client` package under `django-rgd-imagery/client`

banesullivan (approved):
This is looking excellent and moving us in a much better direction! Thanks @AlmightyYakob 

Only one minor suggestion to make sure the workflow to push to PyPI is right

.github/workflows/release.yml
AlmightyYakob: I've made these changes, but I'm assuming you'll want to create the `rgd-imagery-client` package yourself, since I believe you have control of the other PyPI packages.

banesullivan: Great, thanks - I will handle publishing to PyPI after this is merged and we're ready

Issue #425, opened 2021-06-04
STAC Item JSON does not validate


banesullivan:
We should be testing the STAC Item JSON we generate. I tried doing so with the commented code in df7fc2974e35adac0acf5da91ebcac5fbc6651d3 but it fails.

cc @mcovalt

PR #525, opened 2021-09-23
PySTAC validation

mcovalt:
Addresses part of #425

- Adds a link to the collection for an Item which is mandatory.
- The Item `platform` property cannot be `null`, so it is set to "unkown" instead of `null`.

There are still validation errors, but they are getting more-and-more cryptic and harder to address. I believe the latest has an issue with the `bands` property on an Item. This is representation is not managed by us, but by PySTAC itself.


banesullivan (approved):
> There are still validation errors, but they are getting more-and-more cryptic and harder to address. I believe the latest has an issue with the bands property on an Item. This is representation is not managed by us, but by PySTAC itself.

In that case, it sounds like we've done everything we can for now to produce a valid STAC item. Let's fix the merge conflict, get this in, and call #425 done.

PR #527, opened 2021-09-24
makes basemap dropdown dynamic

marySalvi:
Changes the baseMap dropdown to be dynamically filled from geo.osmLayer.tileSources


banesullivan (approved):
Works great, thanks!

PR #528, opened 2021-09-24
Add new composed configuration mixins for downstream projects

### LINKED ISSUE #497: Create optional ConfigMixin for downstream projects
Add a new `ConfigMixin` class for RGD's settings that downstream projects can optionally use. This will make our lives easier in that we won't have to create a minor PR in each downstream project everytime a piece of middleware or some odd setting needs to change.

For example, #492 introduced new middleware (`django-crum`) which had to be added to every downstream project's list of middleware: 

- https://github.com/ResonantGeoData/RD-WATCH/pull/20
- https://github.com/ResonantGeoData/RD-OpenGeo/pull/17
- https://github.com/ResonantGeoData/RD-OASIS/pull/11
- https://github.com/ResonantGeoData/RD-TileServerDemo/pull/3
- And other internal/private code

This was somewhat burdensome and not sustainable moving forward. For example, #281 will likely implement new middleware and we'll have to do this once more. 

To minimize this burden, we can include a `ConfigMixin` for `django-composed-configuration` that downstream projects will import. This will be purely opt-in to make our lives maintaining the 5+ downstream projects more feasible.

This `ConfigMixin` class will contain something like the following:

```py
class ResonantGeoDataConfigMixin(ConfigMixin):
    @staticmethod
    def before_binding(configuration: Type[ComposedConfiguration]):
        configuration.INSTALLED_APPS += ['django.contrib.gis']
        try:
            import re

            import osgeo

            libsdir = os.path.join(
                os.path.dirname(os.path.dirname(osgeo._gdal.__file__)), 'GDAL.libs'
            )
            libs = {
                re.split(r'-|\.', name)[0]: os.path.join(libsdir, name)
                for name in os.listdir(libsdir)
            }
            configuration.GDAL_LIBRARY_PATH = libs['libgdal']
            configuration.GEOS_LIBRARY_PATH = libs['libgeos_c']
        except Exception:
            # TODO: Log that we aren't using the expected GDAL wheel?
            pass

        # Install apps in RGD core's dependencies
        configuration.INSTALLED_APPS += [
            's3_file_field',
        ]

        configuration.REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'].append(
            'rest_framework.authentication.BasicAuthentication'
        )

        configuration.AUTHENTICATION_BACKENDS.insert(0, 'rules.permissions.ObjectPermissionBackend')

        configuration.MIDDLEWARE += ['crum.CurrentRequestUserMiddleware', ]
```

supersedes https://github.com/girder/django-composed-configuration/issues/141

banesullivan:
Resolve #497 

This adds a new module called `configuration` to `django-rgd` which is an extension of `django-composed-configuration` for Geospatial and RGD-based projects. This implementation is in response to https://github.com/girder/django-composed-configuration/issues/141#issuecomment-904730622 with further motivation outlined in #492 

New projects from the cookie-cutter can now just add:

```py
from rgd.configuration import ResonantGeoDataBaseMixin

...

class MyProjectMixin(ResonantGeoDataBaseMixin, ..., ConfigMixin):
    ...
    @staticmethod
    def before_binding(configuration: ComposedConfiguration) -> None:

        # Install additional apps
        configuration.INSTALLED_APPS += [
            'rgd',
            'rgd_3d',
            'rgd_fmv',
            'rgd_geometry',
            'rgd_imagery',
        ]
```

To install RGD to their Project

----

In the event that someone tries to use this module without the proper dependencies, they see the following error:

```py
>>> from rgd.configuration import *
../rgd/configuration/__init__.py:1: in <module>
    from .configuration import GeoDjangoMixin, ResonantGeoDataBaseMixin, SwaggerMixin  # noqa
../rgd/configuration/configuration.py:10: in <module>
    raise ImportError(
E   ImportError: Please install `django-composed-configuration` and `django-configurations` to use the configuration mixins. This can be done through the `configuration` extra when installing `django-rgd`.

### CODE REVIEW

**Comment by banesullivan** [comment]
An alternate approach would be to include these directly into `django-rgd` and add `composed-configuration` as an optional dependency to keep this opt-in rather than default. In retrospect, I'm thinking this may be cleaner. However, non-RGD users may still benefit from `GeoDjangoMixin` and wouldn't want to install `django-rgd` and it's whole stack just to get that mixin.

File: django-rgd/rgd/configuration/configuration.py [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+        engine='django.contrib.gis.db.backends.postgis',
+        conn_max_age=600,
+    )
+
+    CELERY_WORKER_SEND_TASK_EVENTS = True
+
+    RGD_FILE_FIELD_PREFIX = values.Value(default=None)
+    RGD_GLOBAL_READ_ACCESS = values.Value(default=False)
+    RGD_AUTO_APPROVE_SIGN_UP = values.Value(default=False)
+    RGD_STAC_BROWSER_LIMIT = values.Value(default=1000)
```
banesullivan, line 70: technically this is a setting from `django-rgd-imagery` but I think we should just include all RGD settings here

banesullivan:
@brianhelba, do you agree with the name for the module chosen here?: `rgd.configuration`

e.g. `from rgd.configuration import ResonantGeoDataBaseMixin`

PR #529, opened 2021-09-24
Add collection parameter to search

### LINKED ISSUE #520: Collection facet in search filter
Users often want to search for data in one (or several) specific Collections - we should add a Collection parameter to the search filter (and thus endpoints and the search form).

To implement this, I think we should also implement an endpoint that will yield all of the user's available Collections. This new endpoint could be used to dynamically build a checkbox list of collections in the search form.

mvandenburgh:
Closes #520.


mvandenburgh:
Whoops, this only lets you select one `Collection` to filter by but the issue says to allow selecting multiple `Collections`. I'll fix that

mvandenburgh:
@banesullivan this should be all set now. I did end up having to add a new REST endpoint that returns a user's `Collections`; @mcovalt is going to open a PR with a base DRF ViewSet class that can be used to convert existing endpoints over, so in the meantime I implemented it as a regular Django view.

django-rgd/rgd/rest/get.py
```diff
@@ -44,3 +44,14 @@ class GetSpatialAsset(RetrieveAPIView, _PermissionMixin):
...
+
+# TODO: temporary until viewset refactor
+class GetUserCollections(ListAPIView, _PermissionMixin):
+    serializer_class = serializers.CollectionSerializer
+    queryset = models.Collection.objects.all()
+
+    def get_queryset(self):
+        if self.request.user.id is not None:
+            return super().get_queryset().filter(collection_permissions__user=self.request.user)
+        return None
```
banesullivan, line 57: This needs to handle the collections that super users can see but might not be assigned. Further, if the setting `RGD_GLOBAL_READ_ACCESS` is enabled, then any `Collection` with no `CollectionPermission` should return as well.

@mvandenburgh, would you please implement a new method called `get_user_collections()` in `django-rgd/rgd/permissions.py` that will handle this logic then have the endpoint call that?

`get_user_collections()` should look *something* like:

```py
def get_user_collections(user):
    queryset = models.Collection.objects.all()
    # Called outside of view
    if user is None:
        return queryset  # Not sure about this case?
    # Must be logged in
    if not user.is_active or user.is_anonymous:
        return queryset.none()
    # Superusers can see all (not staff users)
    if user.is_superuser:
        return queryset
    # Else, check user permissions
    condition = Q(collection_permissions__user=self.request.user)
    if getattr(settings, 'RGD_GLOBAL_READ_ACCESS', False):
        condition |= Q(**{'collection_permissions__isnull': True})
    return queryset.filter(condition)
```

banesullivan (changes requested):
Looking great! One change needed to make sure the proper collections appear given the user and project settings

django-rgd/rgd/rest/get.py
mvandenburgh: Implemented this in https://github.com/ResonantGeoData/ResonantGeoData/pull/529/commits/898fb7a5c8e406b6dca467038e65731909d1a6d2 with a slight modification.

django-rgd/rgd/filters.py
```diff
@@ -74,6 +76,14 @@ class SpatialEntryFilter(filters.FilterSet):
         method='filter_time_of_day',
     )
 
+    collections = filters.ModelMultipleChoiceFilter(
+        help_text='One or more collections that the data might belong to.',
+        label='Collections',
+        method='filter_collections',
+        queryset=Collection.objects.all(),
+        field_name='files__collection',
```
banesullivan, line 84: Unfortunately, this is not sufficient. Downstream models do not all implement a `files` field and so we have to go through the utilities in `permissions.py` to get the paths to the file resources for each downstream model. This is rather complicated and turned out to be quite a bit of work to get right; which is on me for not foreseeing when assigning this to you.

banesullivan (changes requested):
@mvandenburgh, see comment here and see #532 for a complete implementation of this.

Let's sync up when you have a chance about what I did in #532 and why all of that additional work is necessary.

aashish24:
@mvandenburgh @banesullivan is this still relevant? The last comment posted was September 29th.

banesullivan:
This is still relevant and I do not want to close this just yet. I will come back to this later

banesullivan:
Replaced by #607 and #608

PR #532, opened 2021-09-30
Update collection filter to handle downstream models

banesullivan:


banesullivan:
@mcovalt, would you please take a look at some of these queries and perhaps help optimize them? At the moment, these permissions queries are taking quite some time to execute.

django-rgd/rgd/permissions.py
```diff
@@ -71,13 +109,66 @@ def get_permissions_paths(model: Model, target_model: Model) -> List[str]:
...
+
+def filter_perm(user: User, queryset: QuerySet, role: int):
     """Filter a queryset."""
     if queryset.model == models.SpatialEntry:
         # Return subclasses of SpatialEntry
         queryset = queryset.select_subclasses()
     # Called outside of view
     if user is None:
+        # TODO: I think this is used if a user isn't logged in and hits our endpoints which is a problem
         return queryset
```
banesullivan, line 172: We need to look into this

PR #536, opened 2021-10-04
Update GeoDjangoMixin

AlmightyYakob:
I've noticed a couple of things while using this mixin downstream.

First, if importing `osgeo` fails, the `ImportError` is caught and ignored, subsequently causing the following error

```
django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal", "GDAL",
"gdal3.2.0", "gdal3.1.0", "gdal3.0.0", "gdal2.4.0", "gdal2.3.0", "gdal2.2.0", "gdal2.1.0", "gdal2.0.0").
Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings.
```

It seems that if this import fails, and these two environment variables aren't set, then this failure will always occur. The proposed solution in this PR is to remove the `try/except`, and just let any import error bubble up. Since apparently for RGD the required version of `GDAL` is fairly high, this import structure should be fine. @banesullivan what was the minimum version that was discussed? Another option might be to check for the existence of these environment variables, if they may be configured in other ways, and only throw an exception if both the import fails, and these variables are missing.

Second, it seems that the existing code used to set the two environment variables is essentially taken directly from the top level of the `osgeo` package. Since these two variables are created in that module, I've updated the `GeoDjangoMixin` to just use these existing variables, instead of running the same code again. If we're pinning `GDAL` to a high minimum version, I think leveraging this from the `osgeo` package should be fine, but there might be edge cases I'm unaware of. @manthey I've tagged you for review, since you may know more about this.

### DIFF
--- a/django-rgd/rgd/configuration/configuration.py
+++ b/django-rgd/rgd/configuration/configuration.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 from typing import Type
 
 try:
@@ -18,23 +17,10 @@ class GeoDjangoMixin(ConfigMixin):
     @staticmethod
     def before_binding(configuration: Type[ComposedConfiguration]):
         configuration.INSTALLED_APPS += ['django.contrib.gis']
-        try:
-            import re
+        import osgeo
 
-            import osgeo
-
-            libsdir = os.path.join(
-                os.path.dirname(os.path.dirname(osgeo._gdal.__file__)), 'GDAL.libs'
-            )
-            libs = {
-                re.split(r'-|\.', name)[0]: os.path.join(libsdir, name)
-                for name in os.listdir(libsdir)
-            }
-            configuration.GDAL_LIBRARY_PATH = libs['libgdal']
-            configuration.GEOS_LIBRARY_PATH = libs['libgeos_c']
-        except Exception:
-            # TODO: Log that we aren't using the expected GDAL wheel?
-            pass
+        configuration.GDAL_LIBRARY_PATH = osgeo.GDAL_LIBRARY_PATH
+        configuration.GEOS_LIBRARY_PATH = osgeo.GEOS_LIBRARY_PATH
 
 
 class SwaggerMixin(ConfigMixin):


manthey:
We package some GDAL wheels (see https://github.com/girder/large_image_wheels).  Technically, we don't *have* to use the pre-built wheels.  You could install GDAL some other way and then use some other python GDAL package.  This change *requires* that we use our wheels.  I'm okay with that change, but we should recognize that we've removed one possible option for how GDAL is deployed.

One reason NOT to use our wheels is that our wheels only include permissively licensed packages in GDAL (and then, not quite all of them, as the build process for some hasn't been worth the effort).  If you needed a restrictively licensed module built into GDAL, you would want to built it yourself (in whatever environment that license allows).  Obviously, we could revisit this if and when that occurs.

AlmightyYakob:
> We package some GDAL wheels (see [girder/large_image_wheels](https://github.com/girder/large_image_wheels)). Technically, we don't _have_ to use the pre-built wheels. You could install GDAL some other way and then use some other python GDAL package. This change _requires_ that we use our wheels. I'm okay with that change, but we should recognize that we've removed one possible option for how GDAL is deployed.

I'm unfamiliar. Why does this change requires that we use those wheels? As long as someone is installing a recent enough version of GDAL, wouldn't this still function correctly?

manthey:
Wheels built with `auditwheel` place libraries and binaries in the python site-packages directory.  By default, GeoDjango doesn't look there (only in the default system library paths).  The code that sets `GDAL_LIBRARY_PATH` and `GEOS_LIBRARY_PATH` (both before and after this change) attempts to find these libraries in the site-packages directory.  The `osgeo.GDAL_LIBRARY_PATH` and `osgeo.GEOS_LIBRARY_PATH` values are convenience values that I added to the wheel so you would have to do the more complicated search of the `GDAL.libs` directory (and, since `auditwheel` has changed where these directories are location within site-packages several times, this also allows the user of the wheel to not need to care if that happens again).  These values won't be set if the wheel wasn't used or a different wheel was built.

By guarding this in a try/except, these values will only be set on configuration if GDAL was installed from our wheel.  When GeoDjango starts, it will try to import those libraries from the default system library if the except branch was taken and from the directory within site-packages if the except branch wasn't needed.  Without the guard, it will fail if we aren't using our wheels.

AlmightyYakob:
> Wheels built with `auditwheel` place libraries and binaries in the python site-packages directory. By default, GeoDjango doesn't look there (only in the default system library paths). The code that sets `GDAL_LIBRARY_PATH` and `GEOS_LIBRARY_PATH` (both before and after this change) attempts to find these libraries in the site-packages directory. The `osgeo.GDAL_LIBRARY_PATH` and `osgeo.GEOS_LIBRARY_PATH` values are convenience values that I added to the wheel so you would have to do the more complicated search of the `GDAL.libs` directory (and, since `auditwheel` has changed where these directories are location within site-packages several times, this also allows the user of the wheel to not need to care if that happens again). These values won't be set if the wheel wasn't used or a different wheel was built.
> 
> By guarding this in a try/except, these values will only be set on configuration if GDAL was installed from our wheel. When GeoDjango starts, it will try to import those libraries from the default system library if the except branch was taken and from the directory within site-packages if the except branch wasn't needed. Without the guard, it will fail if we aren't using our wheels.

I'd like to handle the case where the wheel isn't used to install GDAL. Would it be sufficient to just wrap the import in a try/except, and if the except branch is reached, log/print that the expected wheel is not being used?

PR #538, opened 2021-10-06
Support API Tokens in client

### LINKED ISSUE #376: API Tokens for Client API
Add the ability to issue API tokens for use with the Python client

Follow up on the discussion in https://github.com/ResonantGeoData/ResonantGeoData/pull/370

1. use `rest_framework.authtoken` for this: https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication
2. set the client up to look for `~/.rgd/token` or `RGD_API_TOKEN` in the env then prompt the user if invalid 
3. use current username/`getpass()` functionality with `rest_framework.authtoken.views.obtain_auth_token` to create the `~/.rgd/token` file automattically: https://www.django-rest-framework.org/api-guide/authentication/#by-exposing-an-api-endpoint

mvandenburgh:
This PR adds support for auth via API tokens through DRF's `rest_framework.authentication.TokenAuthentication` in the `django-rgd` app.
This slightly changes the auth workflow of the client.

Old workflow:
- The user provides their username and password to the `create_rgd_client` function and directly uses them to set the session's `Authorization` header for Basic Auth.

New workflow:
- If the user has no API key, they can provide a username and password to the `create_rgd_client` function just like before. 
  - Behind the scenes, the client will call the API token endpoint to retrieve an API key for the user and save it in a file `~/.rgd/token`.
- If the user does have an API key, they can either save it in a file `~/.rgd/token` OR set the environment variable `RGD_API_TOKEN` to it.
- The `create_rgd_client` function will use the API key to set the session's `Authorization` header.

Closes #376


banesullivan:
> ... I'm not sure what the exact best approach for user experience would be, but I'd think we'd want some mechanism for changing / logging out users.

I think having a function like `clean_tokens()` that deletes the token file would be sufficient here. Perhaps we should also have a parameter when logging in for the first time that prevents saving the token in the rare case that a user doesn't want their token stored.

Otherwise, I think this will be mostly used by having an environment variable of the API key for running in CI or other pre-configured Python environments. @mvandenburgh, can you make sure that the API key *is not* stored to disk when it is passed as an environment variable?

mvandenburgh:
> > ... I'm not sure what the exact best approach for user experience would be, but I'd think we'd want some mechanism for changing / logging out users.
> 
> I think having a function like `clear_tokens()` that deletes the token file would be sufficient here. Perhaps we should also have a parameter when logging in for the first time that prevents saving the token in the rare case that a user doesn't want their token stored.

Added both of these (`clear_token()` function and `save_api_key_to_disk` parameter to `create_rgd_client`).

> Otherwise, I think this will be mostly used by having an environment variable of the API key for running in CI or other pre-configured Python environments. @mvandenburgh, can you make sure that the API key _is not_ stored to disk when it is passed as an environment variable?

This is the current behavior. If the `RGD_API_TOKEN` is set, the API key will not be stored to disk.

PR #543, opened 2021-10-19
Add base classes for ViewSets

mcovalt:
This PR introduces a base class that further views created across the applications should inherit from. A series of other PR's will follow that expand these base classes and transition current views to use them.


mcovalt:
I've moved the base ViewSet classes to the module you've suggested. Also, in https://github.com/ResonantGeoData/ResonantGeoData/pull/543/commits/fd8a86fe67abaf9081f29c0f09242d4898152a7d, I changed the mixin being used to the one introduced from https://github.com/ResonantGeoData/ResonantGeoData/pull/539.

banesullivan (approved):

PR #551, opened 2021-11-02
Support multi-file images

### LINKED ISSUE #519: File grouping mechanism
Following from https://github.com/ResonantGeoData/RGD-ScrumBoard/issues/38

We need a way to check out multiple files at once in a single directory. The ENVI format is a two file pair and if our files are located remotely via URL then we currently have no way of checking them out in one directory as many of the tools that read them need. (There are other formats similar to this, ENVI is just one example)

This is further complicated by the FUSE utility which requires a `..` suffix on the file paths. This is problematic when we use an external library that expects a specific extension to swap out for another (e.g. `.hdr` for `.raw` in ENVI). The `..` suffix will break any logic in those external libs for swapping the extension.

This is also going to be relevant to the work happening in RD-OASIS as the processing models will need to check out multiple `ChecksumFile`s in a single directory at a time (cc @AlmightyYakob).

----

To proceed with this, I propose that we add an `ancillary_files` field to the `Image` model and adjust the ETL routines to check out/FUSE mount all of those files in one directory when processing. We will extend this logic to other models that already have an `ancillary_files` field.

We will likely have to modify the `simple_httpfs` package to no longer require the suffixed `..` to make this all work. It seems to me that'd we just need to adjust [this section of code](https://github.com/higlass/simple-httpfs/blob/b9195e1fa3888ae256afa7187f488a8aee68b718/simple_httpfs/httpfs.py#L269-L282) and drop support for the `..-wal` and `..-journal` modes. Alternatively, we could open an issue in the `simple-httpfs` repo to see what its devs think about this sort of use case and if there could be support for it (it does not seem terribly active to me though)

### LINKED ISSUE #550: Ensure clean_file_cache does not delete files being used by other processes
In `clean_file_cache`, we should put a minimum limit on the age of files before deleting them.

See https://github.com/ResonantGeoData/ResonantGeoData/blob/c7b82e3d22d075ea4201464b2b7eaab0783f7a42/django-rgd/rgd/utility.py#L284

We should prevent any files less than, say, 5 minutes from being deleted incase they are being used by another process. Otherwise, `clean_file_cache` could easily delete files being used by another process.

----

Perhaps a better approach would be to use the `lockfile` module to lock a directory that is being used such that we know not to delete it.

banesullivan:
Resolve #519, resolve #550

This adds a new `FileSet` model that `ChecksumFile`s can optionally belong to. Think of it as a way of grouping `ChecksumFile` records together. My implementation here preserves "downstream" relationships (e.g. `Image.file = ForiegnKey(ChecksumFile)`) by some trickery when calling `yield_local_path()` on a `ChecksumFile` -- it will download all files in that file's `file_set` before yeilding.

I expanded the work in #548 to download files in the cache directory as a `FileSet` which gives us support for multi-file images. To add a new image with a header file, upload both the data and header files and put them in the same `FileSet`, create an `Image` pointing to the data file, and voila! Please see the test `test_raster_with_header_file` for an example.

Notes:
- This completely removes VSI support with GDAL
- This is a temporary solution until we can implement a custom FUSE interface that will checkout multiple `ChecksumFile`s in one directory.
- This will remain a backup solution for scenarios/environments in which FUSE cannot be/is not available

Issues:
- Two files with different permissions could be checked out under the least restrictive permissions, giving users access to data they might not actually have access to
- `FileSet` currently cannot use the `PermissionPathMixin`

@ResonantGeoData/developers, I'm seeking feedback on:
- [x] The general implementation
- [x] The naming choice (is `Folder` an okay choice? I was also thinking `Item`) - named `FileSet`
- [ ] Should we refactor the permissions to be on `FileSet` to mitigate the issue listed above?
- [x] Should `FileSet` have any fields? e.g. `name`?
- [ ] Should we adjust the `ancillary_files` field on `Raster` to use these new features? Should we make sure all files associated with a raster belong to the same `FileSet`?


django-rgd/rgd/utility.py
```diff
@@ -294,17 +294,34 @@ def clean_file_cache(override_target=None):
...
         path = paths.pop(0)
-        if os.path.isdir(path):
-            shutil.rmtree(path)
-        else:
-            os.remove(path)
+        # Check if resource is locked and skip if so
+        lock_path = Path(f'{path}.lock')
+        lock = FileLock(lock_path)
+        try:
+            with lock.acquire(timeout=0.000000001):
```
manthey, line 301: Filelock's docs state that you probably just want 0 here (https://py-filelock.readthedocs.io/en/latest/api.html#filelock.BaseFileLock.acquire)

mcovalt:
The term "archive" might better describe the model as opposed to "file":

[From Wikipedia](https://github.com/ResonantGeoData/ResonantGeoData/pull/551)
> In computing, an archive file is a computer file that is composed of one or more files along with metadata.

Maybe "VirtualArchive" is a better term?

manthey:
In large_image when a tile source can have multiple files, it is flagged as `_mayHaveAdjacentFiles = True`.

django-rgd/rgd/utility.py
banesullivan: Ah good catch, thanks!

AlmightyYakob:
Noting this from our discussion: Adding a unique constraint on the `name` and `fileset` (or whatever that turns into) fields would allow for `ChecksumFile` names to be unique within a specific `FileSet`, while allowing for overlap globally. It also kind of makes sense with the name `FileSet`, as "set" implies unique members, and it seems a common use case for the grouping of these files is to download them to disk.

mcovalt (approved):
Looks good!

PR #556, opened 2021-11-03
Add missing psutil dep in django-rgd setup.py

AlmightyYakob:
`psutil` is used in `django-rgd`, but not listed as a dependency. Recently, the CI for https://github.com/girder/Danesfield has been failing, due to missing the `psutil` module. I'm not sure what caused this to happen spontaneously, but my best guess is an upstream dep recently removed `psutil` from _it's_ dependencies.

This PR adds `psutil` to `install_requires`.

### DIFF
--- a/django-rgd/setup.py
+++ b/django-rgd/setup.py
@@ -67,6 +67,7 @@ setup(
         'psycopg2',
         'python-magic',
         'flower',
+        'psutil',
         # Production-only
         'django-s3-file-field[minio]',
     ],


banesullivan:
Are you using pre-releases upstream in Danesfield? Also, it is odd that this wasn't caught in our test environments...

banesullivan (approved):

AlmightyYakob:
> Are you using pre-releases downstream in Danesfield? Also, it is odd that this wasn't caught in our test environments...

No, we're just listing `django-rgd[configuration]` as a dependency, and since no pre-release versions are specified, I believe pip will only install full versions.

banesullivan:
Well, that's odd because I've only been making pre-releases lately: https://pypi.org/project/django-rgd/#history

And psutil was introduced in `0.2.6.dev7`

banesullivan:
@AlmightyYakob, should I make another pre-release for this addition?

AlmightyYakob:
> Well, that's odd because I've only been making pre-releases lately: [pypi.org/project/django-rgd/#history](https://pypi.org/project/django-rgd/#history)
> 
> And psutil was introduced in `0.2.6.dev7`

Nevermind, we are installing prereleases in Danesfield, not sure how I missed that.


> @AlmightyYakob, should I make another pre-release for this addition?
I'd say so.

PR #565, opened 2021-11-05
Add support for plugin dependencies

### LINKED ISSUE #564: Client: Sibling plugin access
Currently, if a third party plugin is created as outlined in `PLUGINS.md`, any installed plugin can be access from the client itself. For example, if I installed a plugin with the namespace `foo`, I'd be able to access that plugin and all it's methods with `client.foo.*`. However, this is only an option when accessing the _client_. If I want to access a plugin from within another plugin (the plugin I'm developing), there is currently no way to do that. If another plugin with the namespace `bar` had some methods I wanted to make use of, there's no way I can access those functions naturally.

I'm not sure what the exact solution for this is. I'm currently looking into an approach that passes the client to each plugin instance, which would allow it to access all available plugin namespaces (ex. `self.client.bar.*`), but this feels slightly messy. I'm open to suggestions here.

AlmightyYakob:
Closes #564 

This PR allows for a plugin to access a sibling plugin by defining a _plugin dependency_. In this context, a "sibling plugin" is just any plugin that's installed alongside the plugin in question. While access to all namespaced plugins was already available from the _client_, prior to this change, plugins themselves could not see other plugins.

This has the following breaking change: Plugins must now inherit from `RgdPlugin`. This both helps with plugin discovery, as well as allows for plugin dependencies to be defined.

I'm fairly happy with this solution, but please take a look and lmk if you think this approach is flawed in some way, or if you think there's a better approach.

### DIFF
--- a/django-rgd-imagery/client/rgd_imagery_client/plugin.py
+++ b/django-rgd-imagery/client/rgd_imagery_client/plugin.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import tempfile
 from typing import Dict, Iterable, Iterator, List, Optional, Tuple, Union
 
-from rgd_client.plugin import RGDPlugin
+from rgd_client.plugin import RgdPlugin
 from rgd_client.types import DATETIME_OR_STR_TUPLE, SEARCH_PREDICATE_CHOICE
 from rgd_client.utils import (
     download_checksum_file_to_path,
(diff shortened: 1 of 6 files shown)


banesullivan (approved):
This is really great and helps tremendously on what I am making for RD-WATCH

PR #566, opened 2021-11-05
[WIP] migrate collection to FileSet

banesullivan:
- [ ] Add concrete definitions for all models in `rgd` and `rgd_imagery`
- [ ] Make sure no two images point to the same FileSet
- [ ] Change `ancillary_files` field on `Raster` to be a `manyToManyField` to `FileSet`
    - [ ] make sure this fileset is not used for an Image
- [ ] add annotation on `Image` to point to `FileSet` @mcovalt 
- [ ] Change all relationships to `ChecksumFile` to point to `FileSet` per new definitions (`Image` is the exception and will be well documented why)

Other things to address/start thinking about:

- [ ] How does a user upload data? First, create a `FileSet`, then add `ChecksumFile`s? how do we implement that in the REST endpoints and Python client cleanly? Can we add some server-rendered pages to handle make this easier for users?
- [ ] when getting images from an ImageSet, check each Image's FileSet collection before yielding
  - think of the use case where two users privately uploaded processed imagery to that ImageSet (through the ProcessedImage models)


waxlamp:
TODO (@banesullivan):
- [ ] document conclusions on this idea, how to handle in the future

PR #570, opened 2021-11-05
Add tree action on ChecksumFile ViewSet

AlmightyYakob:
This PR adds a `tree` action, which allows for files to be searched by their name prefix. The result serves to mimic a normal folder hierarchy, returning files/folders at each level. For example, if you had the following files injested:

* `folder/a.txt`
* `folder/b.txt`
* `folder/c/d.txt`

These can be viewed as such

```
folder/
| -- a.txt
| -- b.txt
| -- c/
|    | -- d.txt
```

If you wanted to search for all files under `folder`, you would just pass `folder` to the `tree` endpoint in `path_prefix`. The result would be roughly the following

```
"folders": {
  "c": {
    "size": 123123,
    "num_files": 1,
    "created": "2021-11-05T21:23:41.498195Z",
    "modified": "2021-11-05T21:23:41.498210Z"
  }
},
"files": {
  "a.txt": <ChecksumFile>,
  "b.txt": <ChecksumFile>
}
```

And extending the path_prefix to `folder/c` would yield the final file `d.txt`.

### DIFF
--- a/django-rgd/rgd/models/file.py
+++ b/django-rgd/rgd/models/file.py
@@ -103,6 +103,10 @@ class ChecksumFile(TimeStampedModel, TaskEventMixin, PermissionPathMixin):
     def basename(self):
         return os.path.basename(self.name)
 
+    @property
+    def size(self):
+        return self.file.size
+
     def get_checksum(self):
         """Compute a new checksum without saving it."""
         if self.type == FileSourceType.FILE_FIELD:
--- a/django-rgd/rgd/models/file.py
+++ b/django-rgd/rgd/models/file.py
@@ -105,7 +105,11 @@ class ChecksumFile(TimeStampedModel, TaskEventMixin, PermissionPathMixin):
 
     @property
     def size(self):
-        return self.file.size
+        # Ensure safe check of self.file
+        try:
+            return self.file.size
+        except ValueError:
+            return None
 
     def get_checksum(self):
         """Compute a new checksum without saving it."""
--- a/django-rgd/rgd/rest/viewsets.py
+++ b/django-rgd/rgd/rest/viewsets.py
@@ -1,6 +1,7 @@
 from django.http import HttpResponseRedirect
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import action
+from rest_framework.response import Response
 from rgd import models, serializers
 from rgd.filters import SpatialEntryFilter
 from rgd.rest.base import ModelViewSet, ReadOnlyModelViewSet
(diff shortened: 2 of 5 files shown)


django-rgd/rgd/models/file.py
```diff
@@ -103,6 +103,10 @@ class Meta:
     def basename(self):
         return os.path.basename(self.name)
 
+    @property
+    def size(self):
+        return self.file.size
```
banesullivan, line 108: This will likely raise an error for URL files as `self.file` should be `None`
```suggestion
    @property
    def size(self):
        if self.type == FileSourceType.FILE_FIELD:
            return self.file.size
        # NOTE: we do not have an easy way of getting URL file sizes on the fly
```

banesullivan (commented):
This is looking good. I'd like @mcovalt's review on the implementation

I have a few thoughts on additional scenarios we might want this to handle:
- What about `FileSet`s? These models have a `name` field that could be used as a parent folder of sorts. Should we add an optional field that is `file_set_prefix` to find those as well? Or only find files in `FileSet`s matching that?
- Should we support `regex` with this? I can think of examples where a user might want to find all files matching `(.+)/foo_(\d+).txt` (e.g. `bane/rules/foo_03.txt`)
- Could we also add an option or perhaps a separate method that will return a flat list of all files? Right now, a user would have to iterate over any returned folders. I think there would be utility in having this return all files that match the result, leaving the tree structure embedded in the `name` field.

django-rgd/rgd/models/file.py
AlmightyYakob: Good catch! I forgot about that case, I'm not exactly sure what to do here. It seems to me the two obvious options are:
1. If it's a URL file, return a size of zero. The problem with this is that now we're conveying an incorrect size field, without any other indication that we're unaware of the actual size.
2. If it's a URL file, mark its size, and any folder that contains it, as `unknown`. This would apply if any folder has any URL files, converting its size to `unknown`. The benefit of this is that we're not confidently displaying an incorrect size, but the detriment is that we're also not conveying _any_ information, even if we do have some file fields.

Thinking about this now, there may be a third approach, which would bridge the gap between these two options. We could change the `size` field on folders to `min_size`, to capture any size info on normal file fields. We could also add a `num_url_files` field to the response, to indicate how many files are missing from the `min_size` field. If `num_url_files == 0`, then `min_size` is the actual size of the folder.

@banesullivan Any thoughts on this?

AlmightyYakob:
> * What about `FileSet`s? These models have a `name` field that could be used as a parent folder of sorts. Should we add an optional field that is `file_set_prefix` to find those as well? Or only find files in `FileSet`s matching that?

IMO I don't think the same name/path parsing mechanics should be used in `FileSet`. Since the main purpose of FileSets are to group ChecksumFiles, I think it would make more sense to provide a `file_set` parameter to limit the search to within the given FileSet. 

> * Should we support `regex` with this? I can think of examples where a user might want to find all files matching `(.+)\/foo_(\d+).txt` (e.g. `bane/rules/foo_03.txt`)

That's a good point, although that might make the implementation quite a bit more complicated. I'm okay with doing this, but it might be better done as a follow-up, so any unknowns associated with that change don't delay this PR.

> * Could we also add an option or perhaps a separate method that will return a flat list of all files? Right now, a user would have to iterate over any returned folders. I think there would be utility in having this return all files that match the result, leaving the tree structure embedded in the `name` field.

I think that's something that should go on the regular listing endpoint for ChecksumFiles, as that's really just a `name__startswith` filter.

django-rgd/rgd/models/file.py
banesullivan: I like the `unknown` value for URL files and using `min_size` to represent the knowledge gap if a folder contains a mix between URL files and FieldFiles

banesullivan:
> IMO I don't think the same name/path parsing mechanics should be used in FileSet. Since the main purpose of FileSets are to group ChecksumFiles, I think it would make more sense to provide a file_set parameter to limit the search to within the given FileSet.

Let's worry about `FileSet`s in a separate PR then. I think it'd be useful to search across `FileSet`s whose name match a pattern for files that match another pattern - but this would require the regex support

-----

Re regex: Good point, let's save it for a follow-up PR then.


------

> I think that's something that should go on the regular listing endpoint for ChecksumFiles, as that's really just a name__startswith filter.

Fair enough. I suppose we need to make a new filter for ChecksumFile objects then. This should probably come as a separate PR

banesullivan (approved):

banesullivan (approved):

Issue #563, opened 2021-11-04
Client: API token saving bug on MacOS


AlmightyYakob:
@banesullivan was experiencing the following issue:

When calling `create_rgd_client`, if there isn't already a token stored locally (in `$HOME/.rgd/token`), the client is supposed to make a request to the server to create it, and save it in that file. On MacOS, it seems this might not be occurring. The file doesn't appear to ever be created (notably though, the `.rgd` folder _is_ present). Furthermore, if you try to manually populate that file with your token, it will correctly read it, but the file will then be gone afterwards.

This doesn't actually affect authorization, as it still just fetches the token from the API and stores it in memory, but the storage issue needs to be looked into.

banesullivan:
Thanks for the summary @AlmightyYakob, this is exactly what I was experiencing.

@mvandenburgh, would you please implement a test for the Python client in `django-rgd/tests/test_client.py` that will create a client instance and check that the token is actually saved to the file it should be. 

If that test passes on CI, then we can look into adding a MacOS runner for the tests. If that still passes, then this is likely just an issue with my set up and I will debug from there. Either way, there are two things I'd like to get out of this:

1. A test to validate API tokens are saved to `~/.rgd/token`
2. A MacOS runner for our client's CI tests (unless this proves to be complicated / a maintenance burden). I'm thinking the django server will still run on linux as a docker service and we only run the Python client tests on MacOS.

banesullivan:
Update: this seems to be working after resetting everything. @mvandenburgh, no need to look into a MacOS runner yet - we can do this later.

@mvandenburgh, can you implement something else as a part of this work though? Can you add a check to make sure the token is valid and try to get a new one if not? For example, I regularly use the Python client to log into different RGD deployments hosted on localhost:8000. I have to remember to delete the token or say `save=False` each time since I frequently switch. I'd like some sort of check when fetching the token from `~/.rgd/token` to make sure it works and if not attempt to get a new one.

mvandenburgh:
> @mvandenburgh, can you implement something else as a part of this work though? Can you add a check to make sure the token is valid and try to get a new one if not? For example, I regularly use the Python client to log into different RGD deployments hosted on localhost:8000. I have to remember to delete the token or say `save=False` each time since I frequently switch. I'd like some sort of check when fetching the token from `~/.rgd/token` to make sure it works and if not attempt to get a new one.

@banesullivan I noticed you closed this issue, do you still want this functionality?

[...]

PR #573, opened 2021-11-10
Add test for saving an API key

mvandenburgh:
Adds test that instantiates a new rgd_client with the `save` option set, and then verifies the API key was correctly saved to `~/.rgd/token`.

Addresses #563 partially (still have to look into adding the MacOS CI runner)


banesullivan (approved):
This test looks good, thanks @mvandenburgh

Save the MacOS runner for later - not a prioriry

PR #577, opened 2021-11-11
Retrieve new API key if the saved key is invalid

### LINKED ISSUE #563: Client: API token saving bug on MacOS

mvandenburgh:
When using an API key from `~/.rgd/token`, the client will verify the key is valid by calling a protected endpoint. If it's not, and the user also provided their username and password, the client will attempt to retrieve a new API key from the server and save it at `~/.rgd/token`.

This PR also adds a test for the new functionality and logging to clarify what's going on when an API key is invalid.

Closes #563

### DIFF
--- a/django-rgd/client/rgd_client/client.py
+++ b/django-rgd/client/rgd_client/client.py
@@ -102,6 +102,7 @@ def _read_api_key(api_url: str, username: str = None, password: str = None) -> O
             return api_key
         else:
             logger.error('Provide your username and password next time to fetch a new one.')
+            return None
 
     return api_key
 
--- a/django-rgd/client/rgd_client/client.py
+++ b/django-rgd/client/rgd_client/client.py
@@ -1,4 +1,5 @@
 import getpass
+import logging
 import os
 from typing import List, Optional, Type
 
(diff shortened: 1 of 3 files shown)


django-rgd/client/rgd_client/client.py
```diff
@@ -75,10 +83,28 @@ def _read_api_key() -> Optional[str]:
...
+        if username is not None and password is not None:
+            logger.warning('Attempting to fetch a new API key...')
+            api_key = _get_api_key(api_url, username, password, save=True)
+            if api_key is not None:
+                logger.warning('Succeeded.')
+            return api_key
+        else:
+            logger.error('Provide your username and password next time to fetch a new one.')
+
+    return api_key
```
banesullivan, line 106: shouldn't it return `None` in this case?

banesullivan (changes requested):
```py
>>> client = create_rgd_client(username='my.name@email.com', api_url='http://localhost:8000/api')
API key is invalid.
Provide your username and password next time to fetch a new one.
>>>
```

Can you make it so that it prompts me for my password here?

banesullivan (approved):

PR #579, opened 2021-11-12
Add support for memcached with large_image

### LINKED ISSUE #554: Use memcached with large_image
Right now, large_image is using Python memory for caching per process (large_image logs `Using python for large_image caching` anytime we request tiles). This means the cache is not actually being used when serving tiles as each request comes in through a different process.

We need to configure large_image to use Memcached:

```py
large_image.config.setConfig('cache_backend', 'memcached')
```

This requires a memcached server running somewhere - we may need this in both the django and celery services. My current impression is that different services (on different nodes/machines) cannot share a cache

banesullivan:
Resolve #554 

~Waiting on a new wheel for pylibmc in large_image_wheels~


banesullivan:
Experiencing this issue with pylibmc (what large image wheels is publishing):

```
pylibmc.Error: error 40 from memcached_set: (0x560f07fa6be0) FAILED TO SEND AUTHENTICATION TO SERVER, no mechanism available,  host: memcached:11211 -> libmemcached/sasl.cc:238
```

Note this discussion https://github.com/lericson/pylibmc/issues/246#issuecomment-888914805 though downgrading did not work for me

manthey:
This is probably related to the issues discussed in this repo: https://github.com/memcachier/libmemcached.  I'll try and see if building with that fix resolves this.

PR #580, opened 2021-11-15
Extract ROIs from Raster detail view

banesullivan:
This fixes the utils/endpoints we had for extracting regions from images and adds a new interface on the raster detail view page to select an ROI within the bounds of a raster and directly download that ROI.

Some things to address:

- [ ] Can we add a toast to alert the user that their ROI is outside the bounds of the raster?

If accessing a large dataset remotely, this will not be very RESTful . In the future (with the new SPA frontend), we should have this tool utilize the `ProcessedImage` model and do this asynchronously with polling to see if the job is done.

## Demo

| UI | Produced ROI |
|---|---|
| Screen Shot 2021-11-14 at 6 48 26 PM |  Screen Shot 2021-11-14 at 6 53 27 PM |

### DIFF
--- a/django-rgd-imagery/rgd_imagery/large_image_utilities.py
+++ b/django-rgd-imagery/rgd_imagery/large_image_utilities.py
@@ -6,7 +6,8 @@ import tempfile
 
 import large_image
 from large_image.tilesource import FileTileSource
-from rgd.utility import get_temp_dir
+from large_image_source_gdal import GDALFileTileSource
+from rgd.utility import get_cache_dir
 from rgd_imagery.models import Image
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


django-rgd-imagery/rgd_imagery/large_image_utilities.py
```diff
@@ -42,11 +51,32 @@ def get_region_pixel(
...
+        path = result
+    else:
+        # Write content to temporary file
+        fd, path = tempfile.mkstemp(
+            suffix=f'.{encoding}', prefix='pixelRegion_', dir=str(get_cache_dir())
+        )
+        os.close(fd)
+        path = pathlib.Path(path)
+        with open(path, 'wb') as f:
+            f.write(result)
```
banesullivan, line 79: @manthey, would you please look over this?

mvandenburgh (commented):
The non-`get_region_pixel` code looks good to me.

> Some things to address:
> 
>     * [ ]  Can we add a toast to alert the user that their ROI is outside the bounds of the raster?

Agreed. When I was testing locally, t wasn't immediately clear to me what was happening when I tried to create an ROI outside the bounds.

Another UX issue I noticed is that it's not possible to clear an the ROI drawing off the map after it's created, unless you create a new one.

banesullivan:
> Another UX issue I noticed is that it's not possible to clear an the ROI drawing off the map after it's created, unless you create a new one.

Ah good feedback. I add a new "Clear" button that will do that:

Screen Shot 2021-11-15 at 12 20 40 PM

PR #592, opened 2021-11-16
PointCloud -> Mesh3D: improve support for general 3D meshes

banesullivan:
Improve 3D module to broadly support 3D meshes, not just point clouds (e.g. OBJ files)

Screen Shot 2021-11-16 at 10 30 54 AM

Removes the previous meta model that existed


django-rgd-3d/rgd_3d/tasks/etl.py
```diff
@@ -57,24 +54,27 @@ def _get_readers():
...
-        pc_entry.vtp_data
-    except ChecksumFile.DoesNotExist:
-        pc_entry.vtp_data = ChecksumFile()
-    if not pc_entry.vtp_data.collection:
-        pc_entry.vtp_data.collection = pc_file.file.collection
-    _file_conversion_helper(pc_file.file, pc_entry.vtp_data.file, method, extension='.vtp')
+    if not mesh_3d.vtp_data:
+        mesh_3d.vtp_data = ChecksumFile()
+    if not mesh_3d.vtp_data.collection:
+        mesh_3d.vtp_data.collection = mesh_3d.file.collection
```
mcovalt, line 72: Eventually we ought to move this type of validation out of ETL and into the [`Model.clean()`](https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.clean) method so other tasks can benefit from it. However, this is the pattern we use elsewhere, so I think this is good for now.

mcovalt (approved):

mcovalt (approved):
Looks good on the model change. I made a comment on where I think some of the validation logic should go in the future, but I don't think that should block this from getting merged in now.

PR #600, opened 2021-11-19
POST endpoints for `rgd_geometry`

mvandenburgh:
- Adds serializer + viewset for `GeometryArchive`
- Enable POST/PUT/PATCH/DELETE operations on existing `Geometry` viewset
  - @banesullivan is this correct? That viewset currently inherits from `ReadOnlyModelViewSet` which disables POST operations, but it is listed in #452 as one of the models to have POST support for.


banesullivan (approved):
Perfect, thanks!

PR #601, opened 2021-11-19
POST endpoints for `rgd_fmv`

mvandenburgh:
Currently, `FMVMetaSerializer` is being used for general GET requests to `rgd_fmv`. However, I believe the correct behavior for a POST request is to only allow direct upload of a FMV file; it should not be allowed to modify the `FMVMeta` as that is autogenerated. That is what I implemented in this PR, but @banesullivan please correct me if I am wrong :slightly_smiling_face:


banesullivan (approved):
That's correct, thanks for catching this @mvandenburgh!

PR #603, opened 2021-11-20
Auth token param and ipyleaflet tile veiwer

banesullivan:
These changes enable support for embedding the authentication token as a query parameter. This is for use cases where we want to "pre-sign" a URL for handing off to 3rd party software. Specifically, this is for the use case of generating a pre-signed tile URL for an image so that we can visualize it interactively with `ipyleafelt`.

@mvandenburgh and @AlmightyYakob, would you please review the changes to the authentication and python client?

-------

Here is an example with a new helper method in the `rgd_imagery_client`:

```py
from ipyleaflet import Map, projections, ScaleControl, FullScreenControl
from rgd_client import create_rgd_client

client = create_rgd_client(username='email@kitware.com', api_url='http://localhost:8000/api')

t = client.imagery.get_leaflet_tile_source(18, band=1, palette='matplotlib.Viridis_20', vmin=50, vmax=200)

m = Map(
        center=(37.7249511580583, -122.27230466902257), 
        zoom=9, crs=projections.EPSG3857, 
       )

m.add_layer(t)
m
```

Screen Shot 2021-11-20 at 11 34 08 AM

### DIFF
--- a/django-rgd-imagery/client/rgd_imagery_client/plugin.py
+++ b/django-rgd-imagery/client/rgd_imagery_client/plugin.py
@@ -344,3 +344,68 @@ class ImageryPlugin(RgdPlugin):
 
         r.raise_for_status()
         return r.json()
+
+    def get_leaflet_tile_source(
+        self,
+        image_id: Union[str, int],
+        band: int = None,
+        palette: str = None,
+        vmin: Union[float, int] = None,
+        vmax: Union[float, int] = None,
+        nodata: Union[float, int] = None,
+        **kwargs,
+    ):
+        """Generate an ipyleaflet TileLayer for the given Image.
+
+        Parameters
+        ----------
+        image_id : Union[str, int]
+            The image ID to serve tiles from
+
+        **kwargs
+            All additional keyword arguments are passed to TileLayer
+
+        Return
+        ------
+        ipyleaflet.TileLayer
+
+        """
+        # Safely import ipyleaflet
+        try:
+            from ipyleaflet import TileLayer
+        except ImportError:
+            raise ImportError('Please install `ipyleaflet` and `jupyter`.')
+
+        # Check that the image source is valid and no server errors
+        r = self.session.get(f'image_process/imagery/{image_id}/tiles')
+        r.raise_for_status()
+        # Embed auth token in URL for external use
+        token = self.session.rgd_auth_token
+
+        params = {}
+        if band is not None:
+            params['band'] = band
+        if palette is not None:
+            # TODO: check this value as an incorrect one can lead to server errors
+            #       perhaps we should catch this, server side and ignore bad ones
+            params['palette'] = palette
+        if vmin is not None:
+            params['min'] = vmin
+        if vmax is not None:
+            params['max'] = vmax
+        if nodata is not None:
+            params['nodata'] = nodata
+
+        url = self.session.create_url(
+            f'image_process/imagery/{image_id}/tiles/{{z}}/{{x}}/{{y}}.png?projection=EPSG:3857&auth_token={token}'
+        )
+
+        for k, v in params.items():
+            url += f'&{k}={v}'
+
+        # Set a default attribution let's folks know how awesome RGD is
+        kwargs.setdefault(
+            'attribution',
+            '<a href="https://github.com/ResonantGeoData">Resonant GeoData</a> (Kitware, Inc.)',
+        )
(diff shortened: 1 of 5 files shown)


AlmightyYakob:
@banesullivan and I discussed this offline, and we decided to change this approach to utilize presigned URLs, as including the entire auth token in the URL presents big security vulnerabilities. We came across [this](https://github.com/marcgibbons/drf_signed_auth) package, which seems to implement exactly what we're looking for. If nothing else it's a good starting point.

banesullivan:
To mitigate these concerns, I added a new `SignatureView` to create expiry signatures per user and a new `SignedURLRestViewMixin` that can be added to downstream views which can use this. 

At present, I only enable this functionality for the tiles endpoint and I think it should remain an opt-in feature. That way if someone gets ahold of a signature with malicious intent, they cannot use it to access any POST endpoints.

django-rgd-imagery/tests/test_rest.py
```diff
@@ -83,3 +91,33 @@ def test_create_and_download_cog(admin_api_client, geotiff_image_entry):
...
+    response = client.get(url)
+    response.raise_for_status()
+    # Let the signature expire and assert that we get an accessed denied error
+    time.sleep(3)
+    response = client.get(url)
+    with pytest.raises(requests.HTTPError):
+        response.raise_for_status()
+    assert response.status_code == 401
+    # Reset the TTL now that test is over
+    settings.RGD_SIGNED_URL_TTL = prior
```
mvandenburgh, line 123: Would it be worth using a [`settings` fixture](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings) here? Then you don't have to worry about resetting the TTL at the end of the test, and we wouldn't need to remember to do this for future tests written around this Django setting.

django-rgd/rgd/rest/authentication.py
```diff
@@ -0,0 +1,59 @@
...
+    Extend the TokenAuthentication class to support signed authentication.
+
+    This takes the form of "http://www.example.com/?signature=<key>".
+    """
+
+    def authenticate(self, request):
+        # Check if 'signature' is in the request query params.
+        # Give precedence to 'Authorization' header.
+        param_name = getattr(settings, 'RGD_SIGNED_URL_QUERY_PARAM', 'signature')
+        if param_name in request.query_params and 'HTTP_AUTHORIZATION' not in request.META:
```
AlmightyYakob, line 46: I think it'd be good to add a check here that the request method is `GET`, as this authentication scheme should really only be used for `GET` requests.

django-rgd-imagery/tests/test_rest.py
banesullivan: Thanks! I didn't know about this fixture. I updated other places we were using the settings in tests

mcovalt (approved):
This looks good!

django-rgd/rgd/rest/viewsets.py
```diff
@@ -49,3 +52,13 @@ def footprint(self, request, pk=None):
...
+
+
+class SignatureView(BaseRestViewMixin, views.APIView):
+    """Generate an expirey URL signature."""
+
+    def post(self, request):
+        signer = UserSigner()
+        signature = signer.sign(user=self.request.user)
+        param = getattr(settings, 'RGD_SIGNED_URL_QUERY_PARAM', 'signature')
+        return response.Response({param: signature})
```
AlmightyYakob, line 64: I think it'd be good to include an `expires_at` param here, indicating the unix timestamp this signature expires as. I'll open an issue so it can be addressed separately.

AlmightyYakob: https://github.com/ResonantGeoData/ResonantGeoData/issues/606

django-rgd-imagery/client/rgd_imagery_client/plugin.py
```diff
@@ -344,3 +344,69 @@ def create_processed_image(
...
+        Return
+        ------
+        ipyleaflet.TileLayer
+
+        """
+        # Safely import ipyleaflet
+        try:
+            from ipyleaflet import TileLayer
+        except ImportError:
+            raise ImportError('Please install `ipyleaflet` and `jupyter`.')
```
AlmightyYakob, line 377: `ipyleaflet` isn't listed anywhere in `setup.py` as an requirement. It probably should be, if nothing else under an extra requirement.

PR #607, opened 2021-12-03
Automatic permissions

mcovalt:
Previously, one had to manually specify a path from any object to its relevant collection. This was accomplished via the `PermissionPathMixin` class.

This PR removes the need to manually specify a permissions path. Instead, the permissions paths are found by traversing the graph of model relationships. With this change:
* the authorization logic is simplified and more clearly written.
* the query performance is dramatically increased.

Query performance suffered due to the previous complexity of manually specifying permission paths. This complexity is removed by abstracting away that logic into a more general graph problem. With this abstraction, we can more clearly use a single `INNER JOIN` as opposed to a series of expensive subqueries.


django-rgd/rgd/permissions.py
```diff
@@ -87,22 +137,26 @@ def filter_perm(user, queryset, role):
...
+        is_creator = path.q(created_by=user)
+        conditions |= is_creator
+        if (
+            getattr(settings, 'RGD_GLOBAL_READ_ACCESS', False)
+            and role == models.CollectionPermission.READER
+        ):
+            # A user can read a file by default if it is not in a collection
+            is_not_in_collection = path.q(collection__isnull=True)
+            has_no_owner = path.q(created_by__isnull=True)
+            conditions |= is_not_in_collection & has_no_owner
```
mcovalt, line 153: See comment on this authorization logic in #608. The commit over there, https://github.com/ResonantGeoData/ResonantGeoData/pull/608/commits/a8bce31040866b5f7d441112f38efb698c85421c, changes this behavior.

banesullivan (approved):
This is excellent! I'm going to merge #608 into this, then merge.

Do I completely understand what you did here... not yet... it may be good for you and I to have a run through of all of this

PR #608, opened 2021-12-03
[alternative] Add collection parameter to search

mcovalt:
Alternative solution to #520 

Uses the nifty tools from #607 

Much of the work done in #529 was merged in. This ought to be much more performant due to the work in #607. 

Note that this changes the authorization logic slightly from #607. See https://github.com/ResonantGeoData/ResonantGeoData/commit/a8bce31040866b5f7d441112f38efb698c85421c. This was included since tests were failing now that `ChecksumFile` has a collection assigned to it in the factory (https://github.com/ResonantGeoData/ResonantGeoData/pull/608/commits/7a26fcfac62a88c7eeebdf320fc88d53a606c6c4).


banesullivan (approved):

banesullivan:
> Note that this changes the authorization logic slightly from #607. See a8bce31. This was included since tests were failing now that ChecksumFile has a collection assigned to it in the factory (7a26fcf).

I'm going to have to manually test this after merging

PR #614, opened 2021-12-07
[hotfix] Pin num queries lower for no permissions

mcovalt:
Due to patch to reduce permissions, a test was failing that asserted query count.


mcovalt:
This is incorrect.

PR #617, opened 2021-12-07
Pin django~=3.2

banesullivan:
Temporarily pin django

### DIFF
--- a/django-rgd/setup.py
+++ b/django-rgd/setup.py
@@ -50,7 +50,7 @@ setup(
     install_requires=[
         'boto3',
         'celery',
-        'django==3.2',  # See PR #264: due to this bug: https://code.djangoproject.com/ticket/31910
+        'django~=3.2',  # See PR #264: due to this bug: https://code.djangoproject.com/ticket/31910
         'django-allauth',
         'django-click',
         'django-crum',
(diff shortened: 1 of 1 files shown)


django-rgd/setup.py
```diff
@@ -50,7 +50,7 @@
     install_requires=[
         'boto3',
         'celery',
-        'django>=3.2',  # See PR #264: due to this bug: https://code.djangoproject.com/ticket/31910
+        'django==3.2',  # See PR #264: due to this bug: https://code.djangoproject.com/ticket/31910
```
manthey, line 53: Do you want this to be `~=3.2` to allow future versions of `3.x`?

banesullivan: I don't think there will be any more minor releases now that Django 4 is out. This should still allow patch updates though.

Also, this is temporary, hopefully only for the next few days (famous last words), until we can support Django 4

banesullivan: Changed it though to be fancy with `~=` since I just learned about that yesterday in https://github.com/conda-forge/staged-recipes/pull/17167#discussion_r762512508

mcovalt (approved):
The failing test will be resolved with https://github.com/ResonantGeoData/ResonantGeoData/pull/620.

PR #618, opened 2021-12-07
3D tiles support

### LINKED ISSUE #590: 3D tiles support and basic CesiumJS viewer
Add a new 3D tiles model to the 3D module that leverages the new `FileSet` model and add a basic CesiumJS viewer for it's SSR detail view

Proposal:

- New `Tiles3D` model (better name suggestions?)
    - Point this model to a single `ChecksumFile` - this will be the JSON file and MUST be a part of a `FileSet`. The rest of the data fils in the 3D tiles set must be in that `FileSet`
- Add new `Tiles3DMeta` model that inherits from `SpatialEntry`. Set up an ETL routine on the `Tiles3D` model's save event that will parse the 3D tiles dataset for it's spatial reference and populate the spatial entry

@mcovalt, please review models/relationships here

mvandenburgh:
Resolve #590 

@banesullivan:
- [x] fix 3D coordinate transformations to populate `SpatialEntry`
- [x] Clean up and finalize ETL routines (autopopulate name field, etc)
- [x] Demo data ingestion for 3D tiles data (dragon and Jacksonville data)
- [x] Tests to ensure ETL works and spatial reference is within some tolerance


django-rgd-3d/rgd_3d/tasks/etl.py
```diff
@@ -78,3 +80,15 @@ def read_mesh_3d_file(mesh_3d):
...
+
+
+def read_3d_tiles_tileset_json(tiles_3d: Union[Tiles3D, int]):
+    if not isinstance(tiles_3d, Tiles3D):
+        tiles_3d = Tiles3D.objects.get(pk=tiles_3d)
+    with tiles_3d.json_file.yield_local_path() as path:
+        with open(path, 'r') as f:
+            tileset_json = json.load(f)
+    try:
+        spatial_reference = tileset_json['root']['boundingVolume']['region']
```
mvandenburgh, line 92: @banesullivan `spatial_reference` here contains the coordinates.

Info on the `boundingVolumes` formats found in 3D tiles can be found on page 4 in this document https://github.com/CesiumGS/3d-tiles/blob/main/3d-tiles-reference-card.pdf

banesullivan: I just pushed some work in progress to get this working. Its 90% there. Just need to handle transforms properly when they are given (like in the dragon example)

django-rgd-3d/rgd_3d/tasks/etl.py
```diff
@@ -78,3 +86,86 @@ def read_mesh_3d_file(mesh_3d):
...
+    # Transform from 3DTiles SRID to Database SRID
+    # Convert the coordinates using pyproj to database SRID from source SRID
+    # This might work on proj>=8 and pyproj>=3.1 - large_image_wheels gives us these versions
+    src = pyproj.CRS(f'EPSG:{srid}')
+    dest = pyproj.CRS(f'EPSG:{DB_SRID}')
+    project = pyproj.Transformer.from_crs(src, dest, always_xy=True).transform
+    result = shapely_transform(project, ShapelyPolygon(coords))
+    logger.info(f'the coords: {list(result.exterior.coords)}')
+
+    outline = Polygon(list(result.exterior.coords), srid=DB_SRID)
```
banesullivan, line 165: I'm running into a rather big roadblock here.

The 3D Tiles format provides an affine transform matrix that needs to be applied to the coordinates which are given in a local coordinate system (almost always). Trouble is, I didn't see a clear way to perform this transform with pyproj and resorted to shapely which only works for 2D transforms.

The problem is that 3D tiles explicitly use EPSG 4978/4979 ([reference](https://github.com/CesiumGS/3d-tiles/tree/main/specification#coordinate-reference-system-crs)) which are both earth-centered, earth-fixed (ECEF) reference frames. ECEF reference frames require the elevation datum to properly transform back to EPSG 4326. You can see this issue in [this example I made on epsg.io](http://epsg.io/transform#s_srs=4978&t_srs=4326&x=1213958.8112409&y=-4737648.1118417) where the latitude is always 0:

Screen Shot 2021-12-07 at 6 57 56 PM

There is a post [here](https://gis.stackexchange.com/questions/386070/converting-from-epsg4978-to-epsg4326-using-pyproj-always-produces-0-0-latitude) that mentions this ECEF issue but there is also [this thread](https://issueexplorer.com/issue/geopandas/geopandas/2044) where users report it not being an issue with pyproj 3.1.0 with PROJ 8.1.

In my local configuration, I have pyproj 3.3.0.1 and PROJ 8.2.0 but I'm unable to get this to work and always end up with 0 for the latitude.

@manthey, would you have a chance to look this over?

------

Here is a simpler example to run/test outside of Django/RGD:

```py
from shapely.geometry import Polygon as ShapelyPolygon
import pyproj

from shapely.geometry import Polygon, shape
from shapely.ops import transform

def trans(data, src, dest, always_xy=False):
    src = pyproj.CRS(src)
    dest = pyproj.CRS(dest)
    project = pyproj.Transformer.from_crs(src, dest, always_xy=always_xy).transform
    return transform(project, data)

data = shape(truth)
output = trans(data, 'EPSG:4326', 'EPSG:4978', always_xy=True)
```

```py
>>> list(output.exterior.coords)
[(1213958.8112408777, -4737648.111841724),
 (1213509.9461615593, -4735896.351587741),
 (1216578.6996032828, -4735108.9649644485),
 (1217028.6997835222, -4736860.433972068),
 (1213958.8112408777, -4737648.111841724)]
```

The above is effectively what is handled to us from 3D tiles *after* performing the affine transform.

The trick is getting into a more reasonable SRID for our database:

```py
>>> test = trans(output, 'EPSG:4978', 'EPSG:4326', always_xy=True)
>>> list(test.exterior.coords)
[(-75.62794214702885, 0.0),
 (-75.62794214702885, 0.0),
 (-75.5908126959205, 0.0),
 (-75.5908126959205, 0.0),
 (-75.62794214702885, 0.0)]
```

😕  why is latitude always 0?

django-rgd-3d/rgd_3d/tasks/jobs.py
```diff
@@ -10,3 +10,13 @@ def task_read_mesh_3d_file(pc_file_pk):
 
     pc_file = Mesh3D.objects.get(pk=pc_file_pk)
     helpers._run_with_failure_reason(pc_file, read_mesh_3d_file, pc_file_pk)
+
+
+@shared_task(time_limit=86400)
```
zachmullen, line 15: Minor thing, but I like to use `timedelta`s to express these sorts of values. Lots of libraries already accept them for time duration parameters, but even if they don't, you can just do:

```suggestion
@shared_task(time_limit=timedelta(days=1).total_seconds())
```

Which improves readability

banesullivan: We should save this for a separate PR and change this everywhere... Perhaps by the introduction of a single setting/configuration?

zachmullen: Good idea to make it a separate PR. In my other projects I do tend to make these config settings, but not necessarily because I want to tweak them all the time, more so just because it's a convenient place to dump tunable static values like this.

django-rgd-3d/rgd_3d/tasks/etl.py
banesullivan: Some updates:

- ECEF is a 3D coordinate system.  It doesn't make sense unless you have 3 coordinates.  The X-Y plane slices through the globe at the equator, which is why latitude is always 0 if Z=0.
- There exists this for 3-D geospatial transforms: https://gitlab.kitware.com/mattbrown11/colmap_processing/-/blob/master/src/colmap_processing/geo_conversions.py
E.g., rmat_enu_ecef and rmat_ecef_enu could be used to build 4x4 affine matrix (would need to populate the translation in the 4rth column)
- This VTK code does the conversion (from any CRS to ECEF) in C++ with proj. pyproj should have something similar and should work the other direction if you just revert the order of the CRSs https://gitlab.kitware.com/vtk/vtk/-/blob/master/IO/Cesium3DTiles/TreeInformation.cxx#L382

banesullivan (approved):

PR #620, opened 2021-12-07
Update number of queries

mcovalt:
The tested number of queries went down once all the work was merged together.


mcovalt:
Failing test will be fixed in https://github.com/ResonantGeoData/ResonantGeoData/pull/622.

PR #623, opened 2021-12-08
Upgrade to Django 4

mcovalt:
Upgrades to Django 4 and tests with both Django 3.2+ and Django 4.


mcovalt:
Tracking https://github.com/jazzband/django-oauth-toolkit/issues/1037 for the remaining issue.

banesullivan:
Is this still blocked? What will it take to pick this up?

mcovalt:
Not blocked anymore. It should be good to go if the tests pass.

banesullivan:
The migrations check is failing for Django 4 😕

mcovalt:
@mvandenburgh Thanks for the heads up! Our requirements are unbound for that package so I think the most recent would be used.

@banesullivan It fails without debug information, unfortunately, too :(

PR #633, opened 2021-12-10
Unique contraint on ChecksumFile.url and Collection.name

banesullivan:
Adds a unique constraint on `ChecksumFile.url` and `Collection.name`


banesullivan:
@AlmightyYakob, merge when ready and update the dependency in RD-WATCH to point to the git commit in this repo over in  https://github.com/ResonantGeoData/RD-WATCH/pull/42

django-rgd/client/rgd_client/plugin.py
```diff
@@ -91,8 +91,22 @@ def create_file_from_url(
...
+        if collection is not None:
+            payload['collection'] = collection
+        data = self.session.get('checksum_file', params=payload).json()
+        # TODO: This is absolutely stumping me...
+        if isinstance(data, list) and data:
+            # Test env returns list
+            return data[0]
+        elif isinstance(data, dict) and data['results']:
+            # User env returns dict
+            return data['results'][0]
```
banesullivan, line 105: I cannot figure this out... the testing environment returns a list in that get response whereas my user env returns a dict. There might be a configuration difference between the testing project and the `example_project`.

Look into later... need to merge.

banesullivan: It's definitely a DRF configuration thing in the test project because this worked in https://github.com/ResonantGeoData/RD-WATCH/pull/42

PR #637, opened 2021-12-10
Only get bounds on paginated queryset on home page

mcovalt:
https://github.com/ResonantGeoData/ResonantGeoData/pull/619 changed the bahavior of the homepage by showing all polygons instead of just the paginated ones. This is an expensive calculation for large datasets.

This PR reverts back to the original behavior.


banesullivan (approved):
Thanks!

PR #640, opened 2021-12-13
[draft] Auto-reload `celery` on code changes

mvandenburgh:
Not finished yet; `watchmedo` needs to be configured so it ignores `.py` files in certain directories such as `.tox/` (otherwise launching `tox` causes the celery worker to freeze up because it's being restarted so much).

### DIFF
--- a/dev/celery.Dockerfile
+++ b/dev/celery.Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get update && \
         libglib2.0-0 \
         ffmpeg \
         fuse \
+        python3-watchdog \
         && \
     rm -rf /var/lib/apt/lists/*
 
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -24,8 +24,8 @@ services:
       dockerfile: ./dev/celery.Dockerfile
     command: [
       "watchmedo", "auto-restart",
-      "--pattern", "*.py",
-      "--ignore-patterns", "./.tox/*",
+      "--pattern", "*.py;",
+      "--ignore-patterns", "./.tox/**/*.py",
       "--recursive",
       "--",
       "celery",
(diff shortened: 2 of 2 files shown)


banesullivan:
watchmedo should have an `--ignore-patterns` option that could specify `./tox/*`

mvandenburgh:
Sorry for letting this languish for so long - to my recollection, I was running into problems getting `ignore-patterns` working and so this got pushed off for later, and I haven't come back to it since. I'm planning to give this another shot in the near future.

banesullivan:
@mvandenburgh, any updates on this since the last time you looked into it?

mvandenburgh:
> @mvandenburgh, any updates on this since the last time you looked into it?

No, I haven't looked back into it yet. I'll prioritize this for this week.

mvandenburgh:
I spent some time looking into this again. Unfortunately, I do not think `watchmedo` fulfills our needs here. I've tried using the `--ignore-patterns` argument to tell watchmedo to ignore .py files inside `.tox/` but haven't been successful; there's also these two open issues which imply that the behavior we're trying to achieve here isn't possible -
https://github.com/gorakhargosh/watchdog/issues/798
https://github.com/gorakhargosh/watchdog/issues/212

I'm happy to look further into this, but at that point I think this would essentially turn into a research spike for evaluating other options besides watchdog. @waxlamp @banesullivan should I go ahead and do that?

banesullivan:
Thanks for following up! Let's not pursue this at this time.

Your research thus far is not in vain, though! Looking into `watchmedo` like this is useful as we need this sort of capability across girder4 projects to improve development. Let's keep an eye on those two issues and maybe one day come back to this on another project or within the cookie-cutter.

Closing this PR

PR #646, opened 2021-12-14
Reorder `rgd_image` URLs

mcovalt:
Closes https://github.com/ResonantGeoData/ResonantGeoData/issues/644

URLs are routed greedily. For example, in the following:

```python
router = SimpleRouter(trailing_slash=False)
router.register(r'api/a', SerializerA)
router.register(r'api/a/b', SerializerAB)
```

... a `POST` to `/api/a/b` would actually use `SerializerA` since the route is a full match.

This PR corrects the orders of URLs in the `rgd_image` application to prevent this issue.


banesullivan:
Thank you @mcovalt!! Merging....

PR #648, opened 2021-12-14
Client bug fixes

### LINKED ISSUE #628: Python client issues
- trailing `/` on `api_url` causes failure
- `create_rgd_client` silently failing if `api_url` is bad

@mvandenburgh, would you please look into these

mvandenburgh:
Closes #628.


banesullivan (approved):
Thanks for implementing tests with this!

PR #649, opened 2021-12-14
Python client: add method for downloading checksum files

mvandenburgh:
To avoid having to deal with merge conflicts, I based this on the branch of #648.

### DIFF
--- a/django-rgd/client/rgd_client/plugin.py
+++ b/django-rgd/client/rgd_client/plugin.py
@@ -170,13 +170,15 @@ class CorePlugin(RgdPlugin):
         """
         return self.session.get('checksum_file/tree', params={'path_prefix': path}).json()
 
-    def download_file(self, id: int, download_location: Optional[str] = None) -> Path:
+    def download_file(self, id: int, download_location: Optional[str] = None, use_id=False) -> Path:
         """
         Download a ChecksumFile.
 
         Args:
             id: The id of the ChecksumFile
             download_location: Path to download file to. Defaults to current working directory.
+            use_id: Set to `True` to use the ChecksumFile's ID as the downloaded
+                    file name instead of it's name. Defaults to `False`.
 
         Returns:
             The path of the newly downloaded file
@@ -198,7 +200,10 @@ class CorePlugin(RgdPlugin):
         # Create the download directory if it doesn't exist
         download_location_path.mkdir(parents=True, exist_ok=True)
 
-        download_location_path /= filename
+        if use_id:
+            download_location_path /= str(id)
+        else:
+            download_location_path /= filename
 
         # Save the file
         with open(download_location_path, 'wb') as file:
--- a/django-rgd/client/rgd_client/plugin.py
+++ b/django-rgd/client/rgd_client/plugin.py
@@ -5,7 +5,7 @@ import validators
 
 from .session import RgdClientSession
 from .types import DATETIME_OR_STR_TUPLE, SEARCH_PREDICATE_CHOICE
-from .utils import spatial_search_params
+from .utils import iterate_response_bytes, spatial_search_params
 
 
 class RgdPlugin:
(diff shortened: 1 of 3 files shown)


banesullivan:
@mvandenburgh, would you please take a look at `download_image_file` and `download_raster` to refactor them to use this? You may actually want to move `download_checksum_file_to_path` upstream into `rgd_client`.

@AlmightyYakob, would you please chime in and review?

django-rgd/client/rgd_client/plugin.py
```diff
@@ -118,3 +119,33 @@ def create_file_from_url(
...
+            download_location_path = Path(download_location)
+        else:
+            # Download to current working directory if download_location isn't given
+            download_location_path = Path.cwd()
+
+        # Create the download directory if it doesn't exist
+        download_location_path.mkdir(parents=True, exist_ok=True)
+
+        # Save the file
+        with open(download_location_path / str(id), 'wb') as file:
```
AlmightyYakob, line 147: We should probably attempt to use the ChecksumFile name here, defaulting to the ID if there's no name for that file (or rather, if the name is the empty string, as I don't think that field is nullable).

django-rgd/client/rgd_client/plugin.py
```diff
@@ -118,3 +119,33 @@ def create_file_from_url(
...
+            for chunk in r.iter_content(1 << 20):
```
AlmightyYakob, line 148: Is there a reason for using `1 << 20` (1048576) as the chunk size here? Or is it just convention?

AlmightyYakob:
Left a couple comments otherwise looks good

django-rgd/client/rgd_client/plugin.py
mvandenburgh: Nope, it was just an arbitrary decision to pick 1MB.

django-rgd/client/rgd_client/plugin.py
banesullivan: they are unique per fileset / collection. Could you also implement a `download_file_set` method here too?

AlmightyYakob: > I agree that would be much nicer, but ChecksumFile names aren't unique.

I still think it'd be okay to download with the name, even if they aren't unique, since this method if just for a single file. If someone is individually downloading multiple files with the same name to the same path, imo that's kind of on them. We could also add something like a `use_id` argument (defaulting to False), if someone really wanted to avoid the file names.

django-rgd/client/rgd_client/plugin.py
AlmightyYakob: In that case, maybe just specifying the direct value `1_048_576` and adding a comment to explain that it's 1MB would make things a bit clearer.

django-rgd/client/rgd_client/plugin.py
banesullivan: +1^

django-rgd/client/rgd_client/plugin.py
banesullivan: What is a normal/reasonable value? Is 1MB standard?

AlmightyYakob: I believe I've also used 1MB in the past for things like this, imo it's fine.

django-rgd/client/rgd_client/plugin.py
mvandenburgh: > We should probably attempt to use the ChecksumFile name here, defaulting to the ID if there's no name for that file (or rather, if the name is the empty string, as I don't think that field is nullable).

When I read this initially I didn't look close enough at the code snippet and thought you were saying allow users to query for a ChecksumFile by its name; rereading this I'm realizing that you were referring to the name of the downloaded file. That makes much more sense :)

mvandenburgh:
> @mvandenburgh, would you please take a look at `download_image_file` and `download_raster` to refactor them to use this? You may actually want to move `download_checksum_file_to_path` upstream into `rgd_client`.

@banesullivan to clarify, these methods return the resulting `bytes` from the downloaded files instead of saving them to disk. Are you saying refactor them to save files to disk instead?

django-rgd/client/rgd_client/plugin.py
```diff
@@ -13,6 +14,49 @@ class RgdPlugin:
...
+    def download_checksum_file_to_path(
+        self, id: int, path: Optional[Path] = None, keep_existing=False, use_id=False
+    ):
+        """
+        Download a RGD ChecksumFile to a given path.
+
+        Args:
+            id: The id of the RGD ChecksumFile to download.
+            path: The root path to download this file to.
+            keep_existing: If False, replace files existing on disk.
```
AlmightyYakob, line 26:
```suggestion
            keep_existing: If False, replace files existing on disk.
            use_id: If True, save this file to disk using it's ID, rather than it's name.
```

django-rgd/client/rgd_client/plugin.py
```diff
@@ -13,6 +14,49 @@ class RgdPlugin:
     def __init__(self, session: RgdClientSession):
         self.session = session
 
+    def download_checksum_file_to_path(
+        self, id: int, path: Optional[Path] = None, keep_existing=False, use_id=False
+    ):
```
AlmightyYakob, line 19: Sorry for the confusion, but this method should be placed in `CorePlugin`, not `RgdPlugin`. `RgdPlugin` is the base plugin, used mainly as a base class for other plugins, to aid in plugin discovery. `CorePlugin` is where the core RGD functionality resides.

mvandenburgh: Ah, makes sense. Fixed this

AlmightyYakob:
It looks like the ImageryPlugin needs to be updated since it was importing `download_checksum_file_to_path` from `rgd_client.utils`, but otherwise this looks great!

mvandenburgh:
> It looks like the ImageryPlugin needs to be updated since it was importing `download_checksum_file_to_path` from `rgd_client.utils`, but otherwise this looks great!

Oh now I see why that was a function in `utils.py` before, looks like tests are passing now.

AlmightyYakob (approved):
LGTM! Thanks for this

Issue #629, opened 2021-12-08
Base Python client improvements


banesullivan:
- [x] new method to download any ChecksumFile (@mvandenburgh)
- [x] new search method to utilize ChecksumFile filtering from #570 (@AlmightyYakob)
- [x] client methods for adding and modifying Collections (@mvandenburgh)
- [ ] Basically make sure all API methods from the core app are exposed in the core plugin (@AlmightyYakob)

@AlmightyYakob and @mvandenburgh, do you both want to look into splitting up this work?

AlmightyYakob:
Sure, @mvandenburgh and I can coordinate on this.

banesullivan:
@mvandenburgh and @AlmightyYakob, I need the `create_file_from_url` method on the Python client to act like a "get or create" method to prevent duplicate data.

Please note the new unique constraint on URLs in #633 

I would like this client method to first check if a file exists for the given URL and collection. If it exists, return that. If it doesn't exist, post a new one.

Similarly, I need this implemented for the `post_stac` method on the watch client in https://github.com/ResonantGeoData/RD-WATCH/pull/42

I need this functionality ASAP - can we work to get a small PR just for this first and then finish the rest of this work?

AlmightyYakob:
> I would like this client method to first check if a file exists for the given URL and collection. If it exists, return that. If it doesn't exist, post a new one.

Just an FYI, I think this would require a new endpoint for retrieving a checksum file from a url.

banesullivan:
Specifically, this would be done through a filter on the existing Viewset. no new endpoint

banesullivan:
URL would be a text field on this endpoint:
![Screen Shot 2021-12-10 at 10 29 15 AM](https://user-images.githubusercontent.com/22067021/145616423-26599984-1044-4f2c-aa8e-86f0835a17eb.png)

banesullivan:
One more thing to figure out:

https://github.com/ResonantGeoData/ResonantGeoData/pull/633/files#r767013110

What's the needed configuration? Let's add that to the tests and document somewhere

PR #650, opened 2021-12-14
Add `file_tree_search` client method

AlmightyYakob:
Addresses part of #629

Note: The test for this client method is intentionally brief, as the endpoint itself already has its own tests, and rewriting those for the client would be redundant.

This PR made me consider the following: Should we be modeling the data in python with Dataclasses / Pydantic / etc.? Would it provide value? Or should we just keep things as dicts. I wouldn't address this here, but just a thought. cc @banesullivan

### DIFF
--- a/django-rgd/client/rgd_client/plugin.py
+++ b/django-rgd/client/rgd_client/plugin.py
@@ -135,3 +135,18 @@ class CorePlugin(RgdPlugin):
 
         r.raise_for_status()
         return r.json()
+
+    def file_tree_search(self, path: str = ''):
+        """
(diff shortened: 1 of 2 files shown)


banesullivan:
> Should we be modeling the data in python with Dataclasses / Pydantic / etc.? Would it provide value? Or should we just keep things as dicts. I wouldn't address this here, but just a thought.

It may be worth having a discussion on this and if it would help. Would you open a new issue for this and briefly describe what this might look like

banesullivan (approved):

PR #652, opened 2021-12-15
Fix get_raster(stac=True)

### LINKED ISSUE #641: Regression: STAC endpoint for RasterMeta missing
This client method is not working:

```py
stac = client.imagery.get_raster(result, stac=True)
```

This is because the underlying STAC endpoint changed.

@mcovalt, is it possible to convert a Raster to STAC without using a Collection ID?

banesullivan:
Resolve #641 and update raster endpoint

### DIFF
--- a/django-rgd-imagery/client/rgd_imagery_client/plugin.py
+++ b/django-rgd-imagery/client/rgd_imagery_client/plugin.py
@@ -101,11 +101,14 @@ class ImageryPlugin(RgdPlugin):
         if isinstance(raster_meta_id, dict):
             raster_meta_id = spatial_subentry_id(raster_meta_id)
 
+        r = self.session.get(f'rgd_imagery/raster/{raster_meta_id}').json()
         if stac:
-            r = self.session.get(f'rgd_imagery/raster/{raster_meta_id}/stac')
-        else:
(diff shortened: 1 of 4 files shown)


django-rgd-imagery/client/rgd_imagery_client/plugin.py
```diff
@@ -101,11 +101,14 @@ def get_raster(self, raster_meta_id: Union[str, int, dict], stac: bool = False)
...
+        r = self.session.get(f'rgd_imagery/raster/{raster_meta_id}').json()
         if stac:
-            r = self.session.get(f'rgd_imagery/raster/{raster_meta_id}/stac')
-        else:
-            r = self.session.get(f'rgd_imagery/raster/{raster_meta_id}')
-        return r.json()
+            # Get collection ID - TODO: if this is zero, will fail - could that happen?
+            collection_id = (
+                r['parent_raster']['image_set']['images'][0]['file']['collection'] or 'default'
+            )
```
banesullivan, line 109: @mcovalt, should I check type here instead -- would a ID of 0 ever occur?

mcovalt: It technically could occur, but unlikely since the ID sequence typically starts at 1 and monotonically increases.

Issue #657, opened 2021-12-15
Converting to ESPG from the RasterMeta CRS fails


mcovalt:
The [projection extension in the STAC API fails](https://github.com/ResonantGeoData/ResonantGeoData/blob/66b0b3ed43e8ce74c9f01aad11aa4fb05340cb8d/django-rgd-imagery/rgd_imagery/stac/serializers/item/__init__.py#L72-L77) when given a CRS of `'+init=epsg:None +type=crs'`.

banesullivan:
Actually... where were you seeing this error? can we reproduce this for a specific raster? We need to figure out where `crs` is being not being set correctly

banesullivan:
Leaving open to gether a bit more context

mcovalt:
These are only erroring from the `drop1` collections.

banesullivan:
Aha, I think someone mentioned this morning that the bounds, etc. of the drop1 STAC items are messed up. #638 should suffice. Though we may want to have the STAC ingest in RD-WATCH handle null crs to make sure the value is desired rather than what is assumed in #638

PR #658, opened 2021-12-15
HOTFIX crs for STAC serializer

banesullivan:
Potential solution for  #657


mcovalt:
This is the best we can do. Thanks!

mcovalt (approved):

PR #667, opened 2022-01-17
Fix `yield_checksumfiles` bug

### LINKED ISSUE #666: `yield_checksumfiles` function doesn't handle `ChecksumFiles` with file paths as their `name` correctly
To reproduce, run [this management script](https://gist.github.com/mvandenburgh/6c9b284b6bf2b684b43236fe0da35a93) and then try to view the `Raster` that it creates. The tile requests will fail with 500 errors with errors about wrong file paths in the server log.

I came across this bug when working on Danesfield. On that project we encode the file path of a given `ChecksumFile` inside the `name` column - for example, `foo/bar/foobar.txt` represents a file hierarchy of
```
foo/
  bar/
    foobar.txt
````

[This line](https://github.com/ResonantGeoData/ResonantGeoData/blob/main/django-rgd/rgd/models/file.py#L290) in RGD is incompatible with this approach, since it assumes the ChecksumFile name is a flat file name instead of a file path. For example, if we have a ChecksumFile with the name `foo.tiff`, the current code would work:
```
path = '/tmp/rgd/file_cache/foo.tiff'
with yield_checksumfiles([self], path.parent):
    yield path
```
Note that `path.parent` would be `/tmp/rgd/file_cache/foo.tiff`, which is correct. But if the file name is `a/b/c/foo.tiff`:
```
path = '/tmp/rgd/file_cache/a/b/c/foo.tiff'
with yield_checksumfiles([self], path.parent):
    yield path
```
the call to `path.parent` would evaluate to `/tmp/rgd/file_cache/a/b/c`, which is incorrect and leads to 500 errors since the server can't find the files in that location.

mvandenburgh:
Fixes #666.

Also adds test parameters to existing `yield_checksumfiles` that exercise this edge case.

### DIFF
--- a/django-rgd/rgd/models/file.py
+++ b/django-rgd/rgd/models/file.py
@@ -230,7 +230,7 @@ class ChecksumFile(TimeStampedModel, TaskEventMixin):
                 elif self.type == FileSourceType.URL:
                     return download_url_file_to_local_path(self.url, dest_path)
 
-    def get_cache_path(self):
+    def get_cache_path(self, root: bool = False):
         """Generate a predetermined path in the cache directory.
 
         This will use the associated FileSet's cache path if this resource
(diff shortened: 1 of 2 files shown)


mvandenburgh:
Marking this as draft as it's most likely the wrong approach to fixing this.

mvandenburgh (commented):
It works correctly for me now :+1: LGTM

PR #672, opened 2022-01-19
Pin django-rgd dep in subpackages

banesullivan:
RGD is not stable enough to guarantee cross-version compatibility between the core package and subpackages. For this reason, they currently share the same versions.

To further safeguard this, `django-rgd` should be pinned in the subpackages to guarantee a working installation.

Also, I think/hope this will improve using dependabot as it should pick up on the coupled dependencies so that all of these packages get bumped in a single PR instead of an individual PR for each subpackage

### DIFF
--- a/django-rgd-3d/setup.py
+++ b/django-rgd-3d/setup.py
@@ -45,7 +45,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        'django-rgd',
+        f'django-rgd=={__version__}',
     ],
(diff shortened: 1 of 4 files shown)


banesullivan:
Personally, I have an issue with this approach and it can lead to some headaches in developer installations... however, these are issues we should not have to deal with given our docker development configuration. The bigger priority is shipping stable code and improve our "continuous deployment" of RGD into downstream projects that are funding development here.

PR #679, opened 2022-01-25
Implement STAC Item endpoint for Raster

### LINKED ISSUE #669: Improve raster STAC endpoint
Currently, to retrieve the STAC Item JSON for a raster, you must know the collection ID of the raster. This isn't very user-friendly and we are more than able to add a new endpoint that enables users to request a STAC Item for a raster without knowing its  collection.

For https://github.com/ResonantGeoData/RGD-Vue/issues/82 and #653 

The following demonstrates how a user currently must do this:

https://github.com/ResonantGeoData/ResonantGeoData/blob/d69eda7c06cc42ddce74e4f69493c2b620a454d2/django-rgd-imagery/client/rgd_imagery_client/plugin.py#L105-L110

This is fine for the Python client we are shipping, but the new Vue front-end needs to do this as well and I think it'd be worthwhile to have a dedicated endpoint for this.

Thoughts, @mcovalt?

banesullivan:
Resolve #669

cc @loulawrence, this is for https://github.com/ResonantGeoData/RGD-Vue/issues/82

This also fixes some broken/missing endpoints in imagery

### DIFF
--- a/django-rgd-imagery/rgd_imagery/rest/viewsets.py
+++ b/django-rgd-imagery/rgd_imagery/rest/viewsets.py
@@ -93,6 +93,4 @@ class RasterViewSet(ModelViewSet, TaskEventViewSetMixin):
     )
     @action(detail=True)
     def stac(self, *args, **kwargs):
-        obj = self.get_object()
-        serializer = self.get_serializer()
-        return Response(serializer.to_representation(obj))
+        return ModelViewSet.retrieve(self, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


django-rgd-imagery/rgd_imagery/rest/viewsets.py
```diff
@@ -76,11 +76,22 @@ class RasterViewSet(ModelViewSet, TaskEventViewSetMixin):
...
         return models.Raster.objects.all()
+
+    @swagger_auto_schema(
+        method='GET',
+        operation_summary='Fetch the STAC Item JSON for this raster.',
+    )
+    @action(detail=True)
+    def stac(self, *args, **kwargs):
+        obj = self.get_object()
+        return Response(stac.serializers.ItemSerializer().to_representation(obj))
```
banesullivan, line 97: @mcovalt, is there a better way to do this, since `stac.serializers.ItemSerializer` is handled in `get_serializer_class()`?

mcovalt: You can override the `queryset` and `serializer_class` by passing them in as kwargs to the action decorator. https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

PR #681, opened 2022-01-26
Add `percent_overlap` filter

### LINKED ISSUE #462: Add new overlap parameter in search filter
A user wants to be able to search for SpatialEntries that have x% overlap with the search geometry. We should add this as an additional field on the `SpatialEntryFilter`.

Perhaps call it `percent_overlap` and accept a number between 0 an 1.

mvandenburgh:
@banesullivan the issue specifies "A user wants to be able to search for SpatialEntries that have x% overlap with the search geometry". I implemented this so all SpatialEntries that overlap the search geometry by *at least* x% are shown. I figured that is better than only matching the exact percentage given. Does that make sense, or do we want an exact match?

Closes #462

### DIFF
--- a/django-rgd/rgd/filters.py
+++ b/django-rgd/rgd/filters.py
@@ -1,8 +1,9 @@
 from django.contrib.gis import forms
-from django.contrib.gis.db.models.functions import GeometryDistance
+from django.contrib.gis.db.models.functions import Area, GeometryDistance, Intersection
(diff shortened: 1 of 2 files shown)


banesullivan (approved):
> I figured that is better than only matching the exact percentage given. Does that make sense, or do we want an exact match?

That is exactly what I was thinking! This looks great to me, thank you!

PR #683, opened 2022-01-28
Optimize rgd_imagery admin page queries

mvandenburgh:
I was using the Django admin to debug #666 and I noticed the `Image` and `ImageSet` list pages were taking 3-5 seconds to load. It turns out, for N rows in the image/image set tables, approximately O(2N) SQL queries are being executed on page load. So, I made some optimizations to make debugging less painful. 
With these changes, my `ImageSet` list page went from 204 queries to 5 queries for 71 rows. It looks like there's still some repetitive queries being made, but not enough to have a major impact on page load time.


mcovalt (approved):
Looks great! Thanks for doing this.

There are many places where we run into these N+1 queries since this application has quite a few tables. Any help reducing these is always appreciated.

PR #688, opened 2022-02-28
Collection iteration and PyTorch data loaders

banesullivan:
Adds a few endpoints and new methods to the Python client for data iteration over a collection


banesullivan:
@banesullivan, come back to this and merge the new API features.

banesullivan:
@mvandenburgh, would you please review (ignoring the `django-rgd/client/rgd_client/pytorch.py` file)

django-rgd/rgd/rest/viewsets.py
```diff
@@ -18,6 +18,29 @@ class CollectionViewSet(ModelViewSet):
...
+        return Response(serializers.ChecksumFileSerializer(files[int(index)]).data)
+
+    @swagger_auto_schema(
+        method='GET',
+        operation_summary='Get the number of items in this collection',
+    )
+    @action(detail=True)
+    def len(self, request, pk):
+        collection = models.Collection.objects.get(pk=pk)
+        return Response({'len': collection.checksumfiles.count()})
```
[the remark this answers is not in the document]
banesullivan: It could (and should be).

Though having this extra endpoint makes the interface a little nicer with the PyTorch data loader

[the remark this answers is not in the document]
mvandenburgh: I see. Something about having a `len` endpoint still feels weird to me, but I don't know enough about the PyTorch data loader to make a conclusive argument that it shouldn't be there :shrug:

mvandenburgh (approved):

django-rgd/rgd/rest/viewsets.py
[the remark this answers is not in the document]
banesullivan: in hindsight, I'm going to change this

django-rgd/rgd/rest/viewsets.py
```diff
@@ -18,6 +18,20 @@ class CollectionViewSet(ModelViewSet):
...
+    )
+    @action(
+        detail=True,
+        serializer_class=serializers.ChecksumFileSerializer,
+        url_path=r'item/(?P<index>\d+)',
+    )
+    def item(self, request, pk, index):
+        collection = models.Collection.objects.get(pk=pk)
+        files = collection.checksumfiles.order_by('pk')
+        return Response(serializers.ChecksumFileSerializer(files[int(index)]).data)
```
banesullivan, line 33: Index errors need to be handled here and raise an `APIException`

django-rgd/rgd/rest/viewsets.py
```diff
@@ -18,6 +19,24 @@ class CollectionViewSet(ModelViewSet):
...
+    @action(
+        detail=True,
+        serializer_class=serializers.ChecksumFileSerializer,
+        url_path=r'item/(?P<index>\d+)',
+    )
+    def item(self, request, pk, index):
+        collection = models.Collection.objects.get(pk=pk)
+        files = collection.checksumfiles.order_by('pk')
+        try:
+            instance = files[int(index)]
```
banesullivan, line 35: @mcovalt, can you think of a better or different way to do this?

I'm wondering if `instance = files[int(index)]` is an expensive call - i.e., does it retrieve ALL ChecksumFiles and then grab the index? Or is there some django-magic happening where that's really just an iterator and `__getitem__` is perform a single query for that single object?

banesullivan: Playing around in the shell:

```py
In [1]: collection = Collection.objects.first()
SELECT t.oid,
       typarray
  FROM pg_type t
  JOIN pg_namespace ns
    ON typnamespace = ns.oid
 WHERE typname = 'hstore'

Execution time: 0.001136s [Database: default]
SELECT typarray
  FROM pg_type
 WHERE typname = 'citext'

Execution time: 0.000602s [Database: default]
SELECT "rgd_collection"."id",
       "rgd_collection"."name",
       "rgd_collection"."description"
  FROM "rgd_collection"
 ORDER BY "rgd_collection"."id" ASC
 LIMIT 1

Execution time: 0.001140s [Database: default]

In [2]: ChecksumFile.objects.update(collection=collection)
UPDATE "rgd_checksumfile"
   SET "collection_id" = 1

Execution time: 0.008151s [Database: default]
Out[2]: 21

In [3]: files = collection.checksumfiles.order_by('pk')

In [5]: files[int(3)]
SELECT "rgd_checksumfile"."id",
       "rgd_checksumfile"."created",
       "rgd_checksumfile"."modified",
       "rgd_checksumfile"."failure_reason",
       "rgd_checksumfile"."status",
       "rgd_checksumfile"."name",
       "rgd_checksumfile"."description",
       "rgd_checksumfile"."checksum",
       "rgd_checksumfile"."validate_checksum",
       "rgd_checksumfile"."last_validation",
       "rgd_checksumfile"."collection_id",
       "rgd_checksumfile"."created_by_id",
       "rgd_checksumfile"."file_set_id",
       "rgd_checksumfile"."type",
       "rgd_checksumfile"."file",
       "rgd_checksumfile"."url"
  FROM "rgd_checksumfile"
 WHERE "rgd_checksumfile"."collection_id" = 1
 ORDER BY "rgd_checksumfile"."id" ASC
 LIMIT 1
OFFSET 3

Execution time: 0.002382s [Database: default]
Out[5]: <ChecksumFile: ChecksumFile object (4)>

In [6]:
```

It looks like the query has: ` LIMIT 1 OFFSET 3` so this should be a very inexpensive operation

mcovalt: Nice! Yeah, Django's ORM is pretty great.

django-rgd/rgd/rest/viewsets.py
```diff
@@ -18,6 +19,24 @@ class CollectionViewSet(ModelViewSet):
...
+        except (IndexError, ValueError):
+            raise APIException(f'index {index} not valid or out of range.')
```
mvandenburgh, line 37: When I test this locally I'm seeing this exception come back as a 500 error - is there a way to configure `APIException` to raise a 400 error instead? Or perhaps
```suggestion
            raise Http404(f'index {index} not valid or out of range.')
```
Would be better? An invalid index is strictly a user error and not a server error so I think 400 is more appropriate.

django-rgd/rgd/rest/viewsets.py
```diff
@@ -18,6 +19,24 @@ class CollectionViewSet(ModelViewSet):
...
+        method='GET',
+        operation_summary='Get the file at the given index in this collection. Associated files are orderd by primary key and the provided index is the index in that ordered set. If the files in the collection change, this will not be reproducible.',
+    )
+    @action(
+        detail=True,
+        serializer_class=serializers.ChecksumFileSerializer,
+        url_path=r'item/(?P<index>\d+)',
+    )
+    def item(self, request, pk, index):
+        collection = models.Collection.objects.get(pk=pk)
```
mvandenburgh, line 32:
```suggestion
        collection = get_object_or_404(models.Collection, pk=pk)
```
`Collection.objects.get` raises an exception if the object doesn't exist, so this will return a 500 error if trying to get a Collection that doesn't exist.

mvandenburgh (commented):
Good catch on the index error, I missed that before. I just have some suggestions about HTTP status code semantics, everything else looks good

django-rgd/rgd/rest/viewsets.py
banesullivan: Hm, good catch on the error code. I added a new `APIException` subclass for this in d54ea26 that will 404

banesullivan: Scratch that, `from rest_framework.exceptions import ValidationError` seems like the right exception

banesullivan:
Thanks for the help getting this right, @mvandenburgh!

PR #703, opened 2022-03-17
Return spatial entries in rgd_3d viewsets

mvandenburgh:
On Danesfield we need a way to retrieve the spatial entry of a 3D tiles dataset. I noticed other models such as `Raster` return their corresponding spatial entry for GET requests in their viewset, so I adopted that behavior for the `rgd_3d` viewsets.


mvandenburgh:
For some reason the `Tiles3DMeta.footprint` field is coming back as a string instead of a GeoJSON object... need to look more into this. No review needed at this time

banesullivan:
> For some reason the Tiles3DMeta.footprint field is coming back as a string instead of a GeoJSON object...

Inherit from `SpatialEntryFootprintSerializer` rather than `SpatialEntrySerializer`.

This separation is so that we do not *have to* return footprints when serializing large lists of SpatialEntries as the `footprint` field can blow up in size, while `outline` is rather efficient

banesullivan (approved):

PR #704, opened 2022-03-18
Add `file://` URL schema support

### LINKED ISSUE #702: Add `file://` schema for mounting in filesystem
Can we have a `ChecksumFile` access a local file mounted in the docker container?

banesullivan:
Resolves #702

### DIFF
--- a/django-rgd/tests/test_checksumfile.py
+++ b/django-rgd/tests/test_checksumfile.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from django.db import IntegrityError
 import pytest
 from rgd.datastore import datastore, registry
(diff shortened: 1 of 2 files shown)


django-rgd/tests/test_checksumfile.py
```diff
@@ -37,6 +37,18 @@ def test_create_url():
     assert model.name
 
 
+@pytest.mark.django_db(transaction=True)
+def test_create_url_local_file():
+    model = ChecksumFile()
+    model.type = FileSourceType.URL
+    model.url = 'file://opt/django-project/README.md'
```
mcovalt, line 44: _Technically_, this should have the leading slash.

```suggestion
    model.url = 'file:///opt/django-project/README.md'
```

> Note that when omitting host, the slash is not omitted (while "file:///foo.txt" is valid, "file://foo.txt" is not, although some interpreters manage to handle the latter).

[wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme)

...but, I think the way this is implemented will take both. Also, not sure how important it is to have that exact at this stage, so I wouldn't block this PR.

banesullivan: Ah 🤦🏻‍♂️  - you're absolutely right, good catch

PR #710, opened 2022-03-25
Add Django4 migrations

banesullivan:
I think this is what is needed to land #623 

Simply ran `makemigrations` with Django 4.0 installed


mcovalt (approved):
Looks good! Hopefully that's all we'll need.

PR #720, opened 2022-04-01
Set `DEFAULT_PAGINATION_CLASS` and `PAGE_SIZE` in test app

### LINKED ISSUE #638: checksumfile python client list vs dict issue
See https://github.com/ResonantGeoData/ResonantGeoData/pull/633#discussion_r767013110

mvandenburgh:
Closes #638 

The problem was that the test app didn't have pagination enabled, so it was just returning a flat list instead of a dict of the form
```
{
	"count": 100,
	"next": "http://localhost:8000/api/rgd/checksum_file?limit=100&offset=100",
	"previous": null,
	"results": [....]
}
```


banesullivan (changes requested):
Thanks, @mvandenburgh! Would you please add a note to `django-rgd/client/README.md` about this setting and how it is needed

banesullivan (approved):

PR #722, opened 2022-04-01
Add tests for all python client methods that didn't have one

### LINKED ISSUE #643: Clean up Python client and test
There are a few remnants of STAC support in the Python client which were moved to WATCH - these need to be removed

Additionally, many of the python client methods are failing -- entirely due to the fact that we changed many of the views/endpoints and forgot to update these.

To mitigate these issues, we need to make sure we have a test for every method of the Python client that calls on an endpoint.

mvandenburgh:
Closes #643 

This PR adds tests for all of the python client methods that were missing one. As for the failing methods mentioned in the issue, it seems they have been fixed because all tests are passing and we now have test coverage for every client method.

### DIFF
--- a/django-rgd/tests/test_client.py
+++ b/django-rgd/tests/test_client.py
@@ -221,3 +221,31 @@ def test_file_download_keep_existing(
         checksum_file.id, tmp_path, keep_existing=False, use_id=True
     )
     assert existing_file_content != file_path.read_bytes()
+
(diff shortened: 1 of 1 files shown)


banesullivan (approved):

AlmightyYakob (approved):
LGTM, thanks for this!

PR #724, opened 2022-04-01
Return URL string instead of redirecting for Images with URL-based ChecksumFiles

### LINKED ISSUE #642: ImageViewSet.data redirect error for S3 files
If an Image (it's ChecksumFile) is stored remotely on S3, this endpoint errors:

https://github.com/ResonantGeoData/ResonantGeoData/blob/efb2ab026c68ffa852a399aa122d9b38c320afcc/django-rgd-imagery/rgd_imagery/rest/viewsets.py#L67-L71

We should provide a workaround of some sort as the Python client calls this endpoint expecting data to be returned.

Perhaps for S3 files, we can just pass back the S3 URL and let the user figure it out?

Otherwise, we'd have to proxy the whole file through RGD's web server

mvandenburgh:
Closes #642

### DIFF
--- a/django-rgd-3d/rgd_3d/rest/viewsets.py
+++ b/django-rgd-3d/rgd_3d/rest/viewsets.py
@@ -1,5 +1,3 @@
-from django.db.models.fields.files import FieldFile
-from django.shortcuts import redirect
 from drf_yasg import openapi
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import action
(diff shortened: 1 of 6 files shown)


mvandenburgh:
@banesullivan RE: [your comment on the issue](https://github.com/ResonantGeoData/ResonantGeoData/issues/642#issuecomment-1086057553) - the other RGD apps do indeed have `data` viewset actions, but none of them seem to serve the same purpose as the ones for `rgd` and `rgd_imagery` (i.e. they do not redirect to file downloads, but instead return JSON), so I've only updated the `rgd` and `rgd_imagery` data endpoints, as well as the 3D tiles `file` endpoint.

banesullivan (approved):

PR #739, opened 2022-04-13
Add RGD_OUTPUT_FILE_HANDLER

banesullivan:
This adds a new feature to be able to control how some of RGD's ETL routines save generated/converted data as we may not always want to upload the data to S3FF.

On a downstream project, I need to save the data to a local path on disk rather than in minio/S3FF and this feature enables that.

### DIFF
--- a/django-rgd-3d/rgd_3d/tasks/etl.py
+++ b/django-rgd-3d/rgd_3d/tasks/etl.py
@@ -14,17 +14,19 @@ from rgd_3d.models import Mesh3D, Tiles3D, Tiles3DMeta
 logger = get_task_logger(__name__)
 
 
-def _file_conversion_helper(source, output, method, prefix='', extension='', **kwargs):
+def _file_conversion_helper(
+    source: ChecksumFile, output: ChecksumFile, method: callable, extension: str = '', **kwargs
+):
     workdir = get_temp_dir()
     with tempfile.TemporaryDirectory(dir=workdir) as tmpdir:
         # NOTE: cannot use FUSE with PyntCloud because it checks file extension
         #       in path. Additionally, The `name` field for the file MUST have
         #       the extension.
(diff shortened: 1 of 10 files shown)


django-rgd/rgd/utility.py
```diff
@@ -214,7 +214,9 @@ def patch_internal_presign(f: FieldFile):
 
 
 @contextmanager
-def output_path_helper(filename: str, output: FieldFile):
+def output_path_helper(filename: str, output: ChecksumFile, final_name: str = None):
```
mvandenburgh, line 217:
```suggestion
def output_path_helper(filename: str, output: ChecksumFile, final_name: Optional[str] = None):
```

With `from typing import Optional` added to the top.

django-rgd-3d/rgd_3d/tasks/etl.py
```diff
@@ -14,17 +14,19 @@
 logger = get_task_logger(__name__)
 
 
-def _file_conversion_helper(source, output_field, method, prefix='', extension='', **kwargs):
+def _file_conversion_helper(
+    source: ChecksumFile, output: ChecksumFile, method: callable, extension: str = '', **kwargs
```
mvandenburgh, line 18: I'm not sure if `callable` is the proper type annotation here, [`typing.Callable`](https://docs.python.org/3/library/typing.html#callable) might be the correct one but idk. We don't run type checking on Girder 4 projects in general, so I guess this isn't a big deal if the code runs okay.

mvandenburgh (approved):
Just some nitpicks about type hints, besides that I don't see any glaring issues

django-rgd-3d/rgd_3d/tasks/etl.py
banesullivan: Ah yeah, `callable` is a function, not a type. `typing.Callable` is definitely right, thanks

PR #756, opened 2022-05-02
Patch FMV endpoints and view

### LINKED ISSUE #754: 500 error on FMV Detail View page
I noticed this bug when trying to use the FMV module on Danesfield. I tested it on a fresh RGD instance and it still occured.

To reproduce -
- Run the `rgd_fmv_wasabi` management command to populate your DB with FMV data
- Attempt to navigate to `http://localhost:8000/rgd_fmv/<:id>/` and it will return a 500 error with the error `string indices must be integers
`.

### LINKED ISSUE #755: 500 error on `/rgd_fmv/{spatial_id}` endpoint
I noticed this bug when trying to use the FMV module on Danesfield. I tested it on a fresh RGD instance and it still occurred.

To reproduce -

- Run the `rgd_fmv_wasabi` management command to populate your DB with FMV data
- Hit the `/rgd_fmv/{spatial_id}` API endpoint and a 500 will be returned with an error ```AttributeError at /api/rgd_fmv/1
Got AttributeError when attempting to get a value for field `file` on serializer `FMVSerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `FMVMeta` instance.
Original exception text was: 'FMVMeta' object has no attribute 'file'.```

banesullivan:
Resolve #754 and resolve #755

Fixes the endpoints:
Screen Shot 2022-05-02 at 1 38 41 PM

### DIFF
--- a/django-rgd-fmv/rgd_fmv/urls.py
+++ b/django-rgd-fmv/rgd_fmv/urls.py
@@ -4,7 +4,8 @@ from rgd_fmv import models, views
 from rgd_fmv.rest import viewsets
 
 router = SimpleRouter(trailing_slash=False)
-router.register(r'api/rgd_fmv', viewsets.FMVViewSet)
+router.register(r'api/rgd_fmv', viewsets.FMVMetaViewSet, basename='fmv-meta')
+router.register(r'api/rgd_fmv/model', viewsets.FMVViewSet, basename='fmv')
 
 urlpatterns = [
     # Pages
--- a/django-rgd-fmv/rgd_fmv/rest/viewsets.py
+++ b/django-rgd-fmv/rgd_fmv/rest/viewsets.py
@@ -1,16 +1,17 @@
 from rest_framework.decorators import action
-from rgd.rest.base import ModelViewSet
+from rgd.rest.base import ModelViewSet, ReadOnlyModelViewSet
 from rgd_fmv import models, serializers
 
 
-class FMVViewSet(ModelViewSet):
+class FMVMetaViewSet(ReadOnlyModelViewSet):
     queryset = models.FMVMeta.objects.all()
-
-    def get_serializer_class(self):
-        if self.action in ['get', 'list']:
-            return serializers.FMVMetaSerializer
-        return serializers.FMVSerializer
+    serializer_class = serializers.FMVMetaSerializer
 
     @action(detail=True, serializer_class=serializers.FMVMetaDataSerializer)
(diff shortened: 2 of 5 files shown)


django-rgd-fmv/rgd_fmv/urls.py
```diff
@@ -4,7 +4,8 @@
 from rgd_fmv.rest import viewsets
 
 router = SimpleRouter(trailing_slash=False)
-router.register(r'api/rgd_fmv', viewsets.FMVViewSet)
+router.register(r'api/rgd_fmv', viewsets.FMVMetaViewSet, basename='fmv-meta')
+router.register(r'api/rgd_fmv/model', viewsets.FMVViewSet, basename='fmv')
```
banesullivan, line 8: Probably a better name of this but I'm fine with this

mvandenburgh: I can't think of a better one, seems fine to me :shrug:

django-rgd-imagery/tests/test_project/settings.py
```diff
@@ -8,3 +8,17 @@
...
+
+    from rgd.models import FileSourceType
+
+    # save to path on disk and give `file://` URL to ChecksumFile
+    path = Path('/opt/django-project/.tox/imagery-dump/', name)
+    with open(path, 'wb') as f:
+        f.write(file_handle.read())
+    instance.type = FileSourceType.URL
+    instance.url = f'file://{path}'
+    instance.file = None
```
mvandenburgh, line 24: Was this added by mistake? Or intended to go in the `rgd_fmv` module?

mvandenburgh (commented):
Just wondering about what the `RGD_OUTPUT_FILE_HANDLER` function is for - besides that this all looks good, and I tested it locally and no longer see those errors.

banesullivan:
Oh shoot, that was not supposed to be committed (a local change I had)

banesullivan:
Forced pushed to clean that up

mvandenburgh (approved):
@banesullivan I added a small test that verifies the endpoint response and serializer output are consistent - PTAL. Rest of this LGTM

mvandenburgh:
Looks like tests are failing because of `https://github.com/ResonantGeoData/ResonantGeoData/blob/main/django-rgd/client/rgd_client/utils.py#L13`, since https://www.resonantgeodata.com/api is no longer up.

PR #768, opened 2022-06-03
Support 3D tiles datasets without `transform` property

mvandenburgh:
`transform` is an optional property according to the 3D tiles spec https://github.com/CesiumGS/3d-tiles/tree/main/specification#transforms, so it shouldn't be assumed that it is always present.


banesullivan (approved):
LGTM, assuming you tested it

mvandenburgh:
I tested this with some 3d tile datasets from danesfield, but according to CI one of the datasets from the rgd_3d_tiles demo data is failing tests... I'll take a look

banesullivan (approved):

banesullivan (approved):

